PR #1, opened 2016-04-17
Code clean up.

nekonomicon:


cl_dll/hl/hl_weapons.cpp
```diff
@@ -621,7 +621,7 @@ void HUD_InitClientWeapons( void )
 	// Pass through to engine
 	g_engfuncs.pfnPrecacheEvent		= gEngfuncs.pfnPrecacheEvent;
 	g_engfuncs.pfnRandomFloat		= gEngfuncs.pfnRandomFloat;
-	g_engfuncs.pfnRandomLong		= RandomLong;
+	g_engfuncs.pfnRandomLong		= gEngfuncs.pfnRandomLong;
```
a1batross: There is a different funtion prototypes. Isn't it?

cl_dll/hud_spectator.cpp
```diff
@@ -874,6 +874,9 @@ bool CHudSpectator::ParseOverviewFile( )
 	
 	char *pfile  = NULL;
 
+	memset( filename, 0, 255 );
+	memset( levelname, 0, 255 );
+	memset( token, 0, 1024 );
```
a1batross: It's a C++. Why not to use array initializers?

cl_dll/hl/hl_weapons.cpp
nekonomicon: Nope. It seems Uncle Mike changed it.

cl_dll/hud_spectator.cpp
nekonomicon: Ok. I'm rework it.

PR #3, opened 2016-05-17
VS 2015 Port of "hlsdk for Xash" and Other Fixes

fuzun:
First of all, I tried to not modify anything that did not give a warning or an error. I did not do anything that could break compatibility.
-
This port contains portions from "hlsdk23 msvc2008 port by
Gary_McTaggart" so first, these are the changes that came with that
project:

- Original VC6 DSPs removed.
- 'ARRAYSIZE' macro in 'engine\eiface.h' undefines a macro with the same
name which defined in 'winnt.h'.
- 'HSPRITE' type on the client renamed to 'SpriteHandle_t' due to
'windef.h' which defined same-named type for itself.

And these are the changes made to make this project compatible for VS2015:
- VS2015 properly configured project files added. Release and debug
folders are located at main folder. (ReleaseHL, ReleaseCL)
->xash.sln(root)
-->'client' project
-->'dlls' project
- Full port to VS 2015.
- Gitignore file fixed(typo) and adjusted.
- vec3_origin renamed to vec3_origin_c in C files due to redefinition.
(vec3_origin is not changed for c++ files. But note that vec3_origin_c
is a typedef rather than a standard vector.)
- Some line endings converted to LF from CR LF.
- VGUI files moved to 'unused' because it VGUI support was broken
already. (xash port of sdk does not have all the files needed for vgui,
even though dsp files have them.)
- Some required casts added.
- Partially adopted to C++11.
- Tested for win32 and android. Both works.
- debug.bat and release.bat removed because they are very old and
probably won't work after this commit.


cl_dll/cl_util.h
```diff
@@ -84,16 +84,18 @@ inline struct cvar_s *CVAR_CREATE( const char *cv, const char *val, const int fl
...
 extern float g_hud_text_color[3];
-inline int DrawSetTextColor(float r, float g, float b)
+inline void DrawSetTextColor(float r, float g, float b)
 {
 	if( hud_textmode->value == 1 )
 		g_hud_text_color[0]=r, g_hud_text_color[1] = g, g_hud_text_color[2] = b;
 	else
 		gEngfuncs.pfnDrawSetTextColor( r, g, b );
+
+	return;
```
a1batross: Purpose of this return?

dlls/aflock.cpp
```diff
@@ -229,7 +229,7 @@ void CFlockingFlyer :: Spawn( )
 	
 	pev->frame = 0;
 	pev->nextthink = gpGlobals->time + 0.1;
-	SetThink( &IdleThink );
+	SetThink(&CFlockingFlyer::IdleThink);
```
a1batross: Fixing these things are really, really good idea! :+1:

cl_dll/death.cpp
```diff
@@ -171,7 +171,8 @@ int CHudDeathNotice :: MsgFunc_DeathMsg( const char *pszName, int iSize, void *p
 
 	gHUD.m_Spectator.DeathMessage(victim);
 
-	for ( int i = 0; i < MAX_DEATHNOTICES; i++ )
+	int i;
```
a1batross: And fixing the variable scopes is really nice idea.

fuzun:
These are the things I got noticed so far with the help of 2008 port. But because of time inavailability, I could not revise it.

I will make a new commit tomorrow to finish this port. Meanwhile please seek for bugs and etc. if you intend  to integrate this port. 

Even though vs is still proprietary, it is now

fuzun:
> free and its new versions are really cool.

a1batross:
I tried to review it, but broken line endings is a hell. It's a good that CRLF is converted to LF, but why not to change it everywhere in a separate commit? 

>Original VC6 DSPs removed.

I don't see any point to remove it, but I also don't see why it MSVC6 be supported in our days. 
Anyway, some developers prefer use oldschool MSVC6, while other prefer only latest Visual Studio.

>'ARRAYSIZE' macro in 'engine\eiface.h' undefines a macro with the same name which defined in 'winnt.h'.

Even if ARRAYSIZE have the same meaning, it's better to undefine it and define again. 

>'HSPRITE' type on the client renamed to 'SpriteHandle_t' due to 'windef.h' which defined same-named type for itself.

Nice! :+1: 

>vec3_origin renamed to vec3_origin_c in C files due to redefinition. (vec3_origin is not changed for c++ files. But note that vec3_origin_c is a typedef rather than a standard vector.)

Yeah, because there is even a compatible with float[3] a helper class, it's an error. So why not to have separate vec3_origins for C and C++. =)

>VGUI files moved to 'unused' because it VGUI support was broken already. (xash port of sdk does not have all the files needed for vgui, even though dsp files have them.)

This port don't uses VGUI because Android. It isn't broken.

I don't have any questions anymore and I think it can be merged after a commit split (because commit with thousands changes is a pure evil). Maybe @mittorn will say something.

a1batross:
I merged the previous PR(I forgot about it :(), where is many casts and variable scopes are fixed.

fuzun:
It should not be a problem as I think it is possible to integrate that pr into this tomorrow.

fuzun:
>Even if ARRAYSIZE have the same meaning, it's better to undefine it and define again.
Yeah it actually does that.
>I don't see any point to remove it, but I also don't see why MSVC6 should be supported in our days. 
I removed it because of incompability concerns. Since a lot of stuff have been changed in order to make it compatible for modern compilers (This case it is msvc14) or removed so It might not be backwards compatible to MSVC6.

New commit:
*Integrated @nekonomicon's pull request with these exclusions:
->Uncommented "TraceResult tr;" because it is used and it gave an error
when it was commented.
->Some setthink() and setuse() parameters differ.
->Did not change abs() to fabs macro in C++ files; instead, <cmath>
used.
*Line endings are now fixed to LF.
*Tested for both win32 and android. There is no problem.

nekonomicon:
>*Integrated @nekonomicon's pull request with these exclusions:

```
view.cpp:137:1: warning: '/*' within block comment [-Wcomment]
/*
^
view.cpp:469:8: warning: declaration shadows a local variable [-Wshadow]
                int             i, contents, waterDist, waterEntity;
                                ^
view.cpp:416:9: note: previous declaration is here
        int                             i;
                                        ^
view.cpp:664:7: warning: declaration shadows a local variable [-Wshadow]
                int i;
                    ^
view.cpp:416:9: note: previous declaration is here
        int                             i;
                                        ^
scoreboard.cpp:529:14: warning: declaration shadows a local variable [-Wshadow]
                        for ( int j = 1; j <= m_iNumTeams; j++ )
                                  ^
scoreboard.cpp:488:9: note: previous declaration is here
        int i, j;
               ^

```
>->Did not change abs() to fabs macro in C++ files; instead, cmath
used.

Good idea.
>->Some setthink() and setuse() parameters differ.
>*Tested for both win32 and android. There is no problem.

Try build without -fpermissive.
```
bigmomma.cpp:1201:19: error: must explicitly qualify name of member function
      when taking its address
        pSpit->SetThink( &Animate );
                         ^~~~~~~~
                          CBMortar::
./cbase.h:372:79: note: expanded from macro 'SetThink'
#define SetThink( a ) m_pfnThink = static_cast <void (CBaseEntity::*)(void)> (a)
                                                                              ^
1 error generated.
```

fuzun:
@nekonomicon - Can you check the latest commit? I was not encountered with these warnings because
*-fpermissive was set by default in android makefiles.
*I worked mainly with msvc14 and it did not give these errors. Which is very strange in my opinion.

Also fabs macros are still exist in some c++ files. Should I leave them as they are or should I add <cmath> to these files and convert them to abs? I think there should be a common header file for cpp files, is it feasible to add <cmath> to that file instead of adding it seperately?

nekonomicon:
>Can you check the latest commit? I was not encountered with these warnings because
*-fpermissive is set by default in android makefiles.

Build success now.
-fpermissive for gcc only(I think for fast porting).
I'm use clang compiler which not support this option. It's default compiler for FreeBSD and OS X.

>Also fabs macros are still exist in some c++ files. Should I leave them as they are or should I add cmath to these files and convert them to abs? I think there should be a common header file for cpp files, is it feasible to add cmath to that file instead of adding it seperately?

I think cmath header is needed.

fuzun:
Ok I think tests are done enough on this pr and it is time to do something. I already integrated nekomicon's commit (latest on master branch) into this and I don't have enough time to keep up with upcoming commits. So consider merging this or it will be gone forever.

a1batross:
Tried to merge it, excluding deleted msvc6 project files.

Got many CRLF<->LF conflicts. 

I recommend you close this PR and start it from the beggining. Without changing CRLF->LF or LF->CRLF.

fuzun:
Hmm maybe is it because of MSVC6? Unfortunately I dont have enough time at the moment to do that. Therefore I am closing this pull request.

a1batross:
@fuzun 
no, MSVC6 isn't a problem here. Original HL code is written for it and it can't be broken so easily. :)

Just conflicts.

PR #8, opened 2016-08-06
Add README.md. Some fixes for FreeBSD. Add Makefile for client

MyLittleRobo:
Client Makefile is taken from hlsdk-client

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,27 @@
+# Half-Life SDK for Xash3D
+
+Half-Life SDK for Xash3D with some fixes.
+
+## How to build
+
+### Windows
+
+TODO
+
+### Linux
+
+TODO
+
+### OS X
+
+TODO
+
+### FreeBSD
+
+```
+    cd dlls
+    gmake CXX=clang++ CC=clang
+    cd ../cl_dll
+    gmake CXX=clang++ CC=clang
+```
+
--- a/cl_dll/Makefile
+++ b/cl_dll/Makefile
@@ -1,0 +1,89 @@
+CC=gcc
+CXX=g++
+SRCS+=../dlls/crossbow.cpp
+SRCS+=../dlls/crowbar.cpp
+SRCS+=../dlls/egon.cpp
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -0,0 +1,27 @@
...
+
+### OS X
+
+TODO
+
+### FreeBSD
+
+```
+    cd dlls
+    gmake CXX=clang++ CC=clang
```
nekonomicon: Just use CC?=gcc CXX?=g++ into Makefile.

MyLittleRobo: I'm on FreeBSD, and even with ?= I need to explicitly set CXX to clang++, because it still considers g++ (which does not exist, since executable name now is g++48)

MyLittleRobo: @nekonomicon maybe it would make sense to use cc and c++ aliases instead of gcc and g++. On FreeBSD they refer to clang.

MyLittleRobo:
By the way it looks like nodes.cpp was the last file that used _LINUX macro. It can be removed from Android.mk too.

README.md
nekonomicon: It's strange, may be wrong link for cc?
>because it still considers g++ (which does not exist, since executable name now is g++48)
It's link.

MyLittleRobo: @nekonomicon cc is clang for me.
I don't actually understand how CC in makefile is supposed to be set to clang automatically. There's no even such environment variable by default.

g++ link appeared after I upgraded gcc.

Anyway I would prefer to set compilers explicitly and clang is probably better choice on FreeBSD, that's why I wrote it in README.md

nekonomicon: >There's no even such environment variable by default.

`
cat /usr/share/mk/sys.mk | grep CC
cat /etc/make.conf | grep CC
`
Good practice: gmake CC=${CC}
Because it's maybe changed into /etc/make.conf.
Also, i can upload pmake-compatible makefiles specially for *BSD, if needed.
Just *BSD uses different implementations of pmake by default, not gmake.

PR #11, opened 2016-08-09
Blue Shift

nekonomicon:


MyLittleRobo:
Note: should be merged to FWGS:blueshift, not master. Probably repo admins should create such branch themselves.

a1batross:
It may be added in blueshift directory in master branch, like Valve did it for dmc and ricochet in newer sdk.

MyLittleRobo:
@a1batross I'm not sure this is good idea. It will make applying patches to hl more complicated (must apply same fixes to both source trees, while when it's in separate branch all you need is just a merge). Probably this makes sense for ricochet and dmc, since they are diverse from hl, but Blue shift code is practically the same, so branch looks like a better option.

nekonomicon:
Maybe use macros https://github.com/ptitSeb/halflife/commit/f4b0b36b141a84b150327297177a254fd93963ef ?

fuzun:
Halted?

nekonomicon:
Any suggestions? Separate folder, branch or macros? I can create other PR, if needed.

a1batross:
Merged to blueshift branch.

PR #12, opened 2016-08-16
Visual Studio 2015 Support

fuzun:
- Did not remove msvc6 project files or old batch scripts.
- Only made required changes for vs2015 and for some high level pvs studio warnings.
- PVS studio spotted some functions that can lead overflow. I will look for these at another branch, as this is only for VS2015.
- Had to change HSPRITE. VS2015 did not even like new define macros added recently.


cl_dll/view.cpp
```diff
@@ -78,7 +78,7 @@ extern cvar_t	*cl_forwardspeed;
 extern cvar_t	*chase_active;
 extern cvar_t	*scr_ofsx, *scr_ofsy, *scr_ofsz;
 extern cvar_t	*cl_vsmoothing;
-extern Vector   dead_viewangles;
+/*extern */ Vector   dead_viewangles;
```
a1batross: Are you sure? Then it's definition should be replaced by declaration in ```input_xash3d.cpp```(e.g. add extern key).

I think same fix(as for input_xash3d.cpp) must be applied for inputw32.cpp.

cl_dll/view.cpp
```diff
@@ -78,7 +78,7 @@ extern cvar_t	*cl_forwardspeed;
 extern cvar_t	*chase_active;
 extern cvar_t	*scr_ofsx, *scr_ofsy, *scr_ofsz;
 extern cvar_t	*cl_vsmoothing;
-/*extern */ Vector   dead_viewangles;
+extern  Vector   dead_viewangles;
```
a1batross: Ah, you returned it back. :)

fuzun: I think we need a proper base where we can see changes. I realized that file very lately. Seeking through all of those non descripted commits isn't a good thing to do :)

common/mathlib.h
```diff
@@ -27,7 +27,7 @@ typedef vec_t vec4_t[4];	// x,y,z,w
 
 struct mplane_s;
 
-extern vec3_t vec3_origin;
+extern vec3_t vec3_origin_c;
```
fuzun: Alfred sdk uses some definitions as a workaround I think. But this is really messed up and I did not get the compiler happy without changing it to this.

dlls/explode.cpp
```diff
@@ -164,11 +164,11 @@ void CEnvExplosion::Use( CBaseEntity *pActivator, CBaseEntity *pCaller, USE_TYPE
 	{
 		pev->origin = tr.vecEndPos + ( tr.vecPlaneNormal * ( m_iMagnitude - 24 ) * 0.6 );
 	}
-	else
+	/*else
 	{
-		pev->origin = pev->origin;
-	}
-
+		pev->origin = pev->origin; // ?
```
fuzun: What is this?

cl_dll/in_defs.h
```diff
@@ -17,9 +17,7 @@
 #define	ROLL	2 
 
 #ifdef _WIN32
-#define HSPRITE HSPRITE_win32
```
fuzun: Does not work for VS2015.

mittorn: It does not build with mingw without it

fuzun: I have noticed it while checking latest commits. Since HSPRITE changed to SpriteHandle_t, it is now ok to add that definition.

mittorn:
i'll check this few days later

fuzun:
CMake lists added so there is no need for this.

PR #14, opened 2016-08-21
Remove unneded headers under GPLv3. Upload some headers from original hlsdk 2.3.

nekonomicon:


a1batross:
As this repo is hlsdk-xash3d(Xash3D's HLSDK), everything Xash3D-only features, like extended APIs, must be kept. I don't know how, rewritten or else.

nekonomicon:
Xash3D's HLSDK uses only features from mobility_int.h and physint.h(it seems for props).
I removed headers which uses with engine and menu.
Also, try compare hlsdk headers with quake headers. It's partially identical.

nekonomicon:
Mistake.

mittorn:
@a1batross, why? It is HLSDK, not XashXT. It is better to remove all GPLv3 code.

a1batross:
@mittorn It's can be a base.

Not all features from XashXT are needed for mod. Sometimes it's a two-three
functions, like Render Interface used in cs16-client.

mittorn:
@a1batross It is already a base for mod porting. Just apply some mod as patch for it. So we need to fix all license conflicts.

PR #29, opened 2017-12-03
goldsource mouse input

FreeSlave:
Add a cmake flag to use goldsource compatible mouse input. This flag means that input_goldsource.cpp will be used instead of input_xash3d.cpp.
The resulting client seems to work on both xash3d and goldsource, on Windows and Linux.
Some possible improvements:
1. Probably we could detect if the current engine is xash3d and use input_xash3d code even if the client was built to be goldsource compatible. What is the reliable way to check if the engine is xash3d?
2. Currently I made windows code to not use SDL2 in order to support the old goldsource versions without SDL2. I did not bring back the old windows-specific joystick support though. I don't have a joystick and can't check it anyway.
3. What is the 'official' way to build this sdk with msvc? Because I changed only CMakeLists.txt and use my own script to produce windows binary.

Changes to goldsource input are based on this PR to original halflife repo. https://github.com/ValveSoftware/halflife/pull/1559

This does not solve the issue with black screen on fullscreen (observable at least on Linux gs).


FreeSlave:
Now input backend is chosen depending on the engine running.
One still need to set a flag GOLDSOURCE_SUPPORT for cmake. This is to avoid dependence on SDL2 when unnecessary.

a1batross:
Well, nice! It's finally done. :)

Merging. (OSX GCC job was stuck, but Clang succeeded).

FreeSlave:
@a1batross we also need to bring back the non-SDL joystick support. If you have joystick, you can try.
Also I'm not sure about all this m_bMouseThread thing in PR to original halflife repo, so I left it disabled.
Probably mouse on windows is not worked as intended for now.

PR #30, opened 2017-12-03
Joystick support for non-sdl win32 (untested)

FreeSlave:
Untested because I don't have a joystick.
If anyone has one, please test.

Note: one needs to link to Winmm.lib and add /DGOLDSOURCE_SUPPORT for msvc flags.

### DIFF
--- a/cl_dll/input_goldsource.cpp
+++ b/cl_dll/input_goldsource.cpp
@@ -217,10 +217,25 @@ enum _ControlList
     AxisTurn
 };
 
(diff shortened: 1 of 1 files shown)


FreeSlave:
Seems to work for one tester.

a1batross:
I will try to test it on Windows, after I replace HDD on my laptop, as it seems dying.

a1batross:
Tested. Works.

PR #34, opened 2017-12-06
Use old API to draw scoreboard and motd lines in goldsource

FreeSlave:


mittorn:
Why new api does not work? hlsdk 2.4 seems to use it

FreeSlave:
I did not investigate.

a1batross:
Maybe it was an incompability with clearing Unicode state, nor HLSDK 2.4 nor CS not using it.

PR #37, opened 2017-12-10
A little fixes.

nekonomicon:


a1batross:
I think you can use #pragma once before the include guards, which may result a faster preprocessing. And it works almost everywhere.

FreeSlave:
You reorganized layour of some variables. Will not it break saves?

dlls/effects.cpp
```diff
@@ -379,17 +379,17 @@ class CLightning : public CBeam
 
 	void BeamUpdateVars( void );
 
+	string_t	m_iszStartEntity;
+	string_t	m_iszEndEntity;
+	string_t	m_iszSpriteName;
```
a1batross: Don't break the offsets, Hamsandwich from AMXX relies on it.

dlls/func_break.h
```diff
@@ -82,9 +82,9 @@ class CBreakable : public CBaseDelay
 
 	Materials m_Material;
 	Explosions m_Explosion;
+	string_t m_iszGibModel;
+	string_t m_iszSpawnObject;
```
a1batross: Same here

dlls/func_tank.cpp
```diff
@@ -97,6 +97,9 @@ class CFuncTank : public CBaseEntity
 
 protected:
 	CBasePlayer* m_pController;
+	string_t	m_iszSpriteSmoke;
+	string_t	m_iszSpriteFlash;
+	string_t	m_iszMaster;	// Master entity (game_team_master or multisource)
```
a1batross: And here.

dlls/lights.cpp
```diff
@@ -37,8 +37,8 @@ class CLight : public CPointEntity
 	static TYPEDESCRIPTION m_SaveData[];
 
 private:
+	string_t m_iszPattern;
 	int m_iStyle;
-	int m_iszPattern;
```
a1batross: And here

a1batross:
Well. I will not comment every layout reorganization, just don't move them.

nekonomicon:
Ok.

PR #40, opened 2017-12-18
Ignore dlls, pdb, some cmake generated files and orig files produced …

FreeSlave:
…by some diff programs

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,14 @@
 *.a
 *.framework
 *.exe
+*.dll
+*.pdb
 build/
+CMakeLists.txt.user
+cmake_install.cmake
+*.orig
+*.cbp
+*.dsw
+*.vsxproj
+*.vsproj
+*.sln


a1batross:
You can also add here vsxproj, vsproj, sln, as they are generated.

FreeSlave:
Generated by what? I thought our cmake script does not support msvc now.

a1batross:
There is nothing special in supporting MSVC.

FreeSlave:
Added.
Btw, what do you think of removing some original xash3d sdk leftovers?

a1batross:
Is there any leftovers? It's a hlsdk for xash3d engine in first place.

nekonomicon:
@FreeSlave, Do you mean headers and props? It's useful for modders.

FreeSlave:
I mean .bat, .lst and .dsw files in the repository root. And also gnu.txt (hlsdk is not gpl anyway)

a1batross:
Some headers are GPL licensed(like com_model.h) and have nothing to do with HLSDK. There is just two licenses, although I don't know are they compatible. I'm not a lawyer.

Yes, I think, outdated rubbish can be removed.

PR #47, opened 2018-01-18
Copyright symbol to utf-8

FreeSlave:
QtCreator does not detect encoding automatically and set non-utf8 files to read-only.
Anyway it's better to have a single encoding for all files.
Another option: change all © entries to (c).


a1batross:
I think it would be better to replace all of them to (c). Qt Creator allows to edit ANSI files.

FreeSlave:
QtCreator allows to edit files in many encodings, but user needs to choose one manually if it's not utf-8.
Anyway now it's all ascii.

a1batross:
Nice. Merging, as there is no actual code changes.

PR #52, opened 2018-03-28
Some fixes for FoV, HUD, egon flare and new compilers.

nekonomicon:


dlls/player.cpp
```diff
@@ -3923,7 +3935,7 @@ void CBasePlayer::UpdateClientData( void )
 	if( pev->health != m_iClientHealth )
 	{
 #define clamp( val, min, max ) ( ((val) > (max)) ? (max) : ( ((val) < (min)) ? (min) : (val) ) )
-		int iHealth = clamp( pev->health, 0, 255 ); // make sure that no negative health values are sent
+		int iHealth = clamp( pev->health, 0, pev->max_health ); // make sure that no negative health values are sent
```
a1batross: It's better to keep 255, as pev->health can be often bigger than max_health value. For example, func_door with negative hurt value.

PR #56, opened 2018-04-03
Add mingw support to CMakeLists.txt. Add crosscompiling instructions …

FreeSlave:
…to README.md


FreeSlave:
I'm not sure when _vsnprintf and _snprintf were added to mingw though, so older versions may not work. I tested on 6.3.0

PR #63, opened 2018-07-08
Fix #62 ligthting issue.

### LINKED ISSUE #62: Black models
On map c4a1

Steam Half-Life:
c4a10000

Release build of the current master:
c4a10000

FreeSlave:


a1batross:
Why does it happens? Some compiler bug?

a1batross:
Ah, I see. "v" is defined as int,

PR #65, opened 2018-08-07
Made library functions hidden by default on Mac/Linux

x6herbius:
This resolves an issue where the server library was calling client
library functions in error.


x6herbius:
On a tangentially related note, I noticed that the client library `Cmakelists.txt` includes files from the server's code folder. As part of this pull request I looked as the possibility of moving them into the `game_shared` folder, but it looked like it would be a lot of work to sort out the dependencies on the other server header files. Would this be a feasible thing to do, or is it not worth it?

a1batross:
I thought I already have added this flag, but seems I just forgot to push it. :)

FreeSlave:
> I looked as the possibility of moving them into the game_shared folder

No need for this. Let things stay classic.

PR #67, opened 2018-11-27
Fix goldsource fullscreen when resolution is different from screen's one

FreeSlave:
How to replicate a bug: make the goldsource fullscreen and change resolution to any that's lower than screen resolution. This is how it looks:

screenshot from 2018-11-27 08-52-09

This patch fixes the issue.

### DIFF
--- a/cl_dll/cdll_int.cpp
+++ b/cl_dll/cdll_int.cpp
@@ -212,7 +212,9 @@ void TeamFortressViewport::paintBackground()
 //	int wide, tall;
 //	getParent()->getSize( wide, tall );
 //	setSize( wide, tall );
-	gEngfuncs.VGui_ViewportPaintBackground(HUD_GetRect());
+	int extents[4];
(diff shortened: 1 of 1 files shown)


a1batross:
Does this will work with vguiless client on GS?

a1batross:
Oops. %)

I mean Xash3D with and without vgui.

What will return getParent() if there is no parents of viewport?

FreeSlave:
If there's no root panel, there will be no gViewPort
https://github.com/FWGS/hlsdk-xash3d/blob/master/cl_dll/cdll_int.cpp#L239

It's not really GS-related, rather VGUI-related. We probably need to test it on xash3d with vgui support too. But I have some issue that freezes the camera in Xash3D on lower resolutions, so I can't test it myself (can't even check if xash3d with vgui had this bug before or not).

a1batross:
Runs fine. Merging.

PR #68, opened 2018-12-03
Enable camera rotation on death in Gold Source

FreeSlave:


a1batross:
Can you please change spaces to tabs in separate PR?

PR #71, opened 2019-02-09
Fix hornetgun recharge time on changelevel

FreeSlave:
Note that valve sdk does not have hornetgun recharge time in save/restore at all. Because of this original behavior for hornetgun is to reset its ammo to maximum upon changelevel.

Recharge time was added to save/restore in https://github.com/FWGS/hlsdk-xash3d/commit/740a5a2c31b799dd3f8bd4e5312451489e3cafbe#diff-0fea681efb1fd613feb6d48bb36e21d2
This commit just fixes wrong variable type.

Depending on our policies regarding preserving original save/restore or not we might remove the change made at the first place.

### DIFF
--- a/dlls/weapons.cpp
+++ b/dlls/weapons.cpp
@@ -1617,7 +1617,7 @@ IMPLEMENT_SAVERESTORE( CEgon, CBasePlayerWeapon )
 
 TYPEDESCRIPTION CHgun::m_SaveData[] =
 {
-	DEFINE_FIELD( CHgun, m_flRechargeTime, FIELD_FLOAT ),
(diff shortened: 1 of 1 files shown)


a1batross:
Does adding a new field in save restore breaks compatibility with original saves? Note that starting from Xash3D 0.99 and Xash3D FWGS 0.20 we can load GoldSource saves, so I think, we should keep compatibility on game DLLs also.

mittorn:
if entity failed to restore, it will be spawned only with base class values. For weapons it should be enough to not break the game.

a1batross:
@mittorn so, this change(and previous with extending saverestore table) is harmless for compatibility? Oh, that's good. Merging then?

PR #72, opened 2019-02-12
Fix for legacy node graph reading with XASH_64BIT

FOTMarut:
- Added dlls/nodes_compat.h, which implements two stub classes for reading legacy CGraph and CLink
- Modified CGraph::FLoadGraph in dlls/nodes.cpp to take advantage of them if GRAPH_VERSION != 16 == iVersion
- Slight changes to dlls/nodes.h to be able to test for the value of GRAPH_VERSION in preprocessor directives

### DIFF
--- a/dlls/nodes.cpp
+++ b/dlls/nodes.cpp
@@ -21,6 +21,7 @@
 #include	"cbase.h"
 #include	"monsters.h"
 #include	"nodes.h"
+#include	"nodes_compat.h"
 #include	"animation.h"
 #include	"doors.h"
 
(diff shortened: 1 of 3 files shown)


dlls/nodes.cpp
```diff
@@ -2386,42 +2387,47 @@ int CGraph::FLoadGraph( const char *szMapName )
...
-		}
+	// Read the graph version number
+	//
+	length -= sizeof(int);
+	if( length < 0 )
+		goto ShortFile;
+	iVersion = *(int *) pMemFile;
+	pMemFile += sizeof(int);
 
+	if( iVersion == GRAPH_VERSION || iVersion == 16 )
```
a1batross: I suggest you to make defines:
```
#define GRAPH_VERSION_32 16
#define GRAPH_VERSION_64 (GRAPH_VERSION * 10)
```
And check them depending on platform target, instead of using magic values like 160 or 16.
Or, which is better, create our own nodegraph format without dependence on platform `void*`, `string_t`, `long` sizes, with compatibility mode for GRAPH_VERSION 16(or drop it forever, nodes can be recreated if needed).

nekonomicon:
Why do you need compatibility with GRAPH_VERSION 160?
What if nodes in .pak files?
Also, Unkle Mike's compilers can saves node files into .bsp.
All mods for Half-Life already has nodes with GRAPH_VERSION 16 anyway.
So need support only for GRAPH_VERSION 16.

dlls/nodes.cpp
FOTMarut: Good point about the magic numbers, I did it like that to keep it platform-agnostic. Like, I personally would prefer:
`#define _GRAPH_VERSION_LEGACY	(16)
#ifdef XASH_64BIT
#define	_GRAPH_VERSION			(16 * 10)
#else
#define	_GRAPH_VERSION			(16)
#endif
`

FOTMarut:
I made the changes to replace the magic numbers already, they're pushed on master.

I also created a new branch "new-node-graph" on my fork to work in parallel.
I committed a change there, in all the classes from nodes.h that are written to disk, to swap all their integer fields with their platform-independent counterparts; I also created some typedef's to sort out the mess and make it more easy to move things around.
It is a retro-compatible commit, meaning that it shouldn't break anything (the actual types are kept the same, for now)

a1batross:
For me this is fine now. 

@mittorn @nekonomicon

PR #74, opened 2019-04-11
Use line height from screen info in motd and scoreboard

FreeSlave:
Text does look right for me, at least on Linux, probably because of different font sizes on different platforms. Better use gHUD.m_scrinfo.iCharHeight for y-pos calculations.

Before:
crossfire0000
crossfire0001

After:
crossfire0003
crossfire0002


a1batross:
Yeah, I applied same fix in cs16client before, but never ported it to hlsdk-xash3d. Nice!

PR #75, opened 2019-04-11
Fix potential crash / data corruption in hud message

FreeSlave:
We can also increase the size of `line`. What do you think?

### DIFF
--- a/cl_dll/message.cpp
+++ b/cl_dll/message.cpp
@@ -240,7 +240,7 @@ void CHudMessage::MessageDrawScan( client_textmessage_t *pMessage, float time )
 {
 	int i, j, length, width;
 	const char *pText;
-	unsigned char line[80];
+	const char *pLineStart;
 
 	pText = pMessage->pMessage;
 	// Count lines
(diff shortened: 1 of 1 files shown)


a1batross:
80 bytes is small buffer. With use of colorcodes, it can be exceeded too fast. 

Maybe this can be rewritten without using an intermediate buffer? Like, well...

```
m_parms.lineLength = 0;
m_parms.width = 0;

while( *pText && *pText != '\n' )
{
	unsigned char c = *pText;

	if (m_parms.lineLength < SOME_LIMIT?? )
	{
		m_parms.width += gHUD.m_scrinfo.charWidths[c];
		m_parms.lineLength++;
	}
	pText++;
}
unsigned char *line = pText - m_parms.lineLength;
pText++;		// Skip LF

m_parms.x = XPosition( pMessage->x, m_parms.width, m_parms.totalWidth );

for( j = 0; j < m_parms.lineLength; j++ )
{
	m_parms.text = line[j];
	int next = m_parms.x + gHUD.m_scrinfo.charWidths[m_parms.text];
	MessageScanNextChar();

	if( m_parms.x >= 0 && m_parms.y >= 0 && next <= ScreenWidth )
		TextMessageDrawChar( m_parms.x, m_parms.y, m_parms.text, m_parms.r, m_parms.g, m_parms.b );
	m_parms.x = next;
}

m_parms.y += gHUD.m_scrinfo.iCharHeight;
```

FreeSlave:
You're right. Looks like we don't need a buffer. We also don't need a limit as long as a passed message is valid (null-terminated). For custom messages we're limited to the size of g_pCustomText.

PR #81, opened 2019-05-27
Add deny noise for func_recharger and func_healthcharger, when activator's health/armor is full

tyabus:


a1batross:
It's from these bugs, which makes Half-Life is so Half-Life.

tyabus:
> It's from these bugs, which makes Half-Life is so Half-Life.
I think this is not a bug, that is a enchancement

a1batross:
>I think this is not a bug, that is a enchancement

Too noticable enchancements should be in mods.

tyabus:
> > I think this is not a bug, that is a enchancement
> 
> Too noticable enchancements should be in mods.

this is a joke?

nekonomicon:
Nope. You can place any enhancements under macro or use cvars.

tyabus:
> Nope. You can place any enhancements under macro or use cvars.

I don't want to put this under macro/cvar, because this enhancement is very tiny. Can you pull this without any cvars/macros?

nekonomicon:
bhop, weapon bobbing and selfgauss fix has one string changes too and was put under cvars.

FreeSlave:
We are mostly fixing bugs in the master branch, not make enhancements. And if you ask people if the original behavior of chargers is a bug or not, opinions will be different. I think It's a matter of taste.
By the way `MAX_NORMAL_BATTERY` and `pActivator->pev->max_health` exist to avoid magic numbers.
There're a lot of things that can be enhanced in Half-Life but better keep them in mods.

tyabus:
oops, i haven't done putting this under cvar

nekonomicon:
@tyabus, you can use the following commands to edit the latest commit:
```git add <changed files>
git commit --amend```

a1batross:
@nekonomicon 

and `git push --force`. Force push is literally a bad thing, but may be good to keep history clean. 

But I think, commits must be atomic and change only one thing.

tyabus:
Done.

PR #84, opened 2019-06-11
replace gnu.txt with LICENSE from ValveSoftware/halflife

iZarif:


nekonomicon:
@iZarif gnu.txt was in archive with original xash3d.
HLSDK licensed under Valve's EULA.
https://github.com/ValveSoftware/halflife/blob/master/LICENSE
https://github.com/FWGS/hlsdk-xash3d/blob/master/dlls/cbase.h#L1

iZarif:
Maybe we should just delete this file?

nekonomicon:
Yes

a1batross:
Maybe adding Half-Life SDK license file instead?

PR #87, opened 2019-07-08
Fix rapid crowbar bug

FreeSlave:
This fixes a rapid crowbar bug (that happens when you hit a dead corpse).

You don't need to merge it into master. Just leave it as PR or make another branch. It can be useful for modders.

@nekonomicon you can use it for your reimplementations of mods.


FreeSlave:
@nekonomicon maybe merge this and another crowbar PR at least into Blue Shift? We already build it with crowbar idle animations enabled.

nekonomicon:
Why not use a macro?

FreeSlave:
Added macros in both PRs

PR #88, opened 2019-07-08
Fix crowbar delay after the first hit

FreeSlave:
Sometimes there's a noticable delay before the next animation after the first hit on breakable.

You don't need to merge it into master. Just leave it as PR or make another branch. It can be useful for modders.

@nekonomicon you can use it for your reimplementations of mods.


nekonomicon:
Why not use a macro?
>@nekonomicon you can use it for your reimplementations of mods.

Many old mods based on HLSDK 2.1 and not use weapon prediction.

PR #90, opened 2019-07-30
Stop gauss pulse sound on overcharge

FreeSlave:
This stops gauss pulse sound from playing after the overcharge.

It's probably better not to merge it to master as it changes the code of both client and server. It can be useful in mods though.


FreeSlave:
Force pushed, adding a macro check.

PR #91, opened 2019-08-02
Opfor update

FreeSlave:
@nekonomicon 

This actually needs testing. I'm just commiting what I've done.

TODO:
* Pitworm is weird and needs rework
* Opfor grunts should have a separate relationship class
* Test everything in the actual game


FreeSlave:
I guess it's playable now, at least in singleplayer.

PR #98, opened 2019-10-02
fix gcc 8.3.0 format-overflow  warnings

iZarif:


nekonomicon:
All good, but size of arrays in cache-friendly code must be power of two.

iZarif:
fixed

PR #104, opened 2019-10-30
Build: Fixed issues on Windows

x6herbius:
in_camera.o wouldn't link because of unresolved
symbols relating to getting/setting the cursor
position, so user32.lib was added as a dependency.
The gitignore was also missing ignore rules for
different waf folder name variations, and the
.vscode directory for people (like me) who use
VS Code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,9 @@ cmake_install.cmake
 *.vsproj
 *.sln
 .waf-*
+waf-*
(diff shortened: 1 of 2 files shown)


a1batross:
Nice, gonna merge it. But can you duplicate changes for CMake as well?

x6herbius:
Just adding now.

PR #106, opened 2019-10-31
Build: Fixed GiveFnptrsToDll not being found on Windows

x6herbius:
Also fixed server DLL being named server.dll instead of hl.dll.


x6herbius:
> 
> 
> This is not valid for gcc

That's a very good point. I should probably change the condition to check the compiler rather than the target platform.

PR #107, opened 2019-10-31
Build: Swapped Win32 check for MSVC check

x6herbius:
@mittorn commented that technically I should have been checking against the compiler rather than the target OS, since `/def` is a MSVC-specific flag.


a1batross:
Oops, didn't noticed it. 

Should we also have same flag for CMake as well?

PR #108, opened 2019-10-31
build: fix travis ci

Aimless-Wanderer:
I think travis is still using Python 3.4.*, so f-strings are not supported yet and android engine doesn't build.

### DIFF
--- a/dlls/wscript
+++ b/dlls/wscript
@@ -16,7 +16,7 @@ def configure(conf):
 		hlDefNode = conf.path.find_resource("./hl.def")
 
 		if hlDefNode is not None:
-			conf.env.append_unique('LINKFLAGS', '/def:{}'.format(hlDefNode.abspath()))
+			conf.env.append_unique('LINKFLAGS', '/def:%s' % hlDefNode.abspath())
 		else:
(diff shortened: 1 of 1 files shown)


dlls/wscript
```diff
@@ -16,7 +16,7 @@ def configure(conf):
 		hlDefNode = conf.path.find_resource("./hl.def")
 
 		if hlDefNode is not None:
-			conf.env.append_unique('LINKFLAGS', f'/def:{hlDefNode.abspath()}')
+			conf.env.append_unique('LINKFLAGS', '/def:{}'.format(hlDefNode.abspath()))
```
a1batross: Let's then use % notation, i.e. `'/def:%s' % hlDefNode.abspath()`, it's shorter and easier to read, IMO.

Aimless-Wanderer: As you wish, I'm not a huge fan of C style formatting in Python :)

a1batross:
Thanks!

dlls/wscript
nekonomicon: Fan or not fan, but even python programmers on my main job uses this style.

PR #114, opened 2019-11-18
WinRT port

MoeMod:


mittorn:
Why not use engine input interface for WinRT instead? It does support mouse (if was not, it was bug in new engine, but it is fixed already)

MoeMod:
> Why not use engine input interface for WinRT instead? It does support mouse (if was not, it was bug in new engine, but it is fixed already)

I have tried but SDL can't restrict the cursor in the window.

mittorn:
Why not fix it on engine side?

mittorn:
https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input.c#L200
https://github.com/FWGS/xash3d-fwgs/blob/master/engine/client/input.c#L576
You may add mouse getting functions somewhere in platform/winrt/ and call it here

hiIexistlol:
pull request [#470](https://github.com/FWGS/xash3d-fwgs/pull/470) on the engine has some mouse reworks in progress

PR #116, opened 2019-12-08
wscript: fix join method usage

Aimless-Wanderer:
Lists don't have a join() method. Strings have.


nekonomicon:
Lists has append method.

Aimless-Wanderer:
But in this case list needs to be converted to a string.

PR #136, opened 2020-09-13
client: wscript: add winmm.lib dependency (win32)

AgentAgrimar:
This fixes a linker error when compiling for win32 with Goldsrc Support enabled.

This is based on a linker failure I experienced on my end, I'm unsure if it's fine for anybody else.

### DIFF
--- a/cl_dll/CMakeLists.txt
+++ b/cl_dll/CMakeLists.txt
@@ -132,6 +132,9 @@ endif()
 
 if(WIN32)
 	target_link_libraries( ${CLDLL_LIBRARY} user32.lib )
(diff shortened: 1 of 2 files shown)


a1batross:
Could you also add it to CMakeLists then? We're trying to keep old cmake as well, since there is people still using it.

Otherwise lgtm.

AgentAgrimar:
> Could you also add it to CMakeLists then? We're trying to keep old cmake as well, since there is people still using it.
> 
> Otherwise lgtm.

Oh woops, guess I forgot about the CMakeLists, I'll do that real quick.

PR #137, opened 2020-09-13
server: fix monster yaw speed

AgentAgrimar:
This makes the yaw speed for all monsters (scientists, headcrabs, etc.) framerate independent.

Fix by [Solokiller](https://github.com/Solokiller) with [mikela-valve](https://github.com/mikela-valve)'s adjustments:
https://github.com/ValveSoftware/halflife/issues/2458

NOTE: **MONSTER_YAWSPEED_FIX** will need to be defined, feel free to remove the need for it if it should be enabled by default.

### DIFF
--- a/dlls/basemonster.h
+++ b/dlls/basemonster.h
@@ -328,5 +328,9 @@ public:
 	BOOL CineCleanup();
 
 	CBaseEntity* DropItem( const char *pszItemName, const Vector &vecPos, const Vector &vecAng );// drop an item.
+
+#ifdef MONSTER_YAWSPEED_FIX
+	float m_flLastYawTime;
+#endif
 };
 #endif // BASEMONSTER_H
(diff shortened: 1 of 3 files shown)


lewa-j:
LGTM

a1batross:
@h1naria can this potentially break compatibility with mods? If not, then we may enable it for everything rather than have it under a macro.

h1naria:
I'm not sure if such code breaks something.
And I think It would be better to place this code under cvar.

h1naria:
Hmm... Wait, I forgot about amxmodx and save/restore.
It may be better to define new fields into monsters classes.

AgentAgrimar:
> Hmm... Wait, I forgot about amxmodx and save/restore.
> It may be better to define new fields into monsters classes.

The way I implemented should be compatible with saves, as I did what mikela-valve did and that's not save/restoring it at all, although looking back at it, I'm not sure if CBaseMonster::MonsterInit() is called at all after a restore, so I might've accidently made it so it's unitialized for one instance of that function being called.

As for amxmodx, although I'm not familiar with that at all, Half-Life itself currently has this change iirc.

h1naria:
@AgentAgrimar I'm about changed offsets to fields in child classes.
But It's really may be not a trouble.

h1naria:
I think need to place this code under cvar:
https://github.com/alliedmodders/amxmodx/pull/771

PR #139, opened 2020-09-23
Fixed wscript files so that waf can be used to compile opfor.

sedme0:


dlls/wscript
```diff
@@ -27,6 +27,21 @@ def build(bld):
 	])
 
 	source += [
+		'gearbox/blkop_apache.cpp', 'gearbox/blkop_osprey.cpp', 'gearbox/cleansuit_scientist.cpp',
```
a1batross, line 30: Check if you can do bld.path.ant_glob(gearbox/*.cpp) instead.

I didn't used glob here because some files are just sample code.

h1naria:
Need to merge changes from master at first.

wscript
```diff
@@ -184,7 +184,7 @@ def configure(conf):
 	compiler_optional_flags = [
 		'-fdiagnostics-color=always',
 		'-Werror=return-type',
-		'-Werror=parentheses',
+		#'-Werror=parentheses',
```
h1naria, line 187: Didn't comment this without reason.

FreeSlave:
Btw in cmake I made ropes.cpp compiling with -O0 (without optimizations). With optimizations it has a weird glitch when a player can randomly teleport between barnacle and rope if both are present on the map. So either we should fix the issue or make the same hack for waf.

nekonomicon:
@FreeSlave I didn't understand why memset was used in this code, and there another code style.

FreeSlave:
I guess it was fixed in https://github.com/FWGS/hlsdk-xash3d/commit/805fce2350179d1f11c8ff65232571d866fb3072
Can we close this PR?

a1batross:
Ye, right.

PR #143, opened 2020-12-25
Fix checking for gag spawnflag

FreeSlave:
Should check against `pev->spawnflags` rather than `pev->flags`


malortie:
@FreeSlave After searching on the official Half-Life repository, I haven't found any issue regarding this problem.

I would recommend to open an issue or PR on the official repo, so that others may know.

nekonomicon:
@malortie this trouble (and some other) was fixed in sven-coop before this PR.

PR #157, opened 2021-02-24
Add Swift Package Manager support

colemancda:
For macOS, iOS and tvOS compilation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ waf3-*
 .lock*
 *.pyc
 .vscode/
+.swiftpm/
+.build
+Package.resolved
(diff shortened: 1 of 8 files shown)


nekonomicon:
1. Don't include all files.
Use file lists from waf/cmake scripts.
2. Don't use `LINUX` or similar macro.
As an example mp_stubs.cpp does not contain windows specific code, bit this file contains stubs for multiplayer-only mods.
3. It would be good to use waf instead of any alternave build system.
4. https://github.com/FWGS/xash3d-fwgs/issues/61

a1batross:
I doubt we would add something than Waf, CMake and Android.mk. Even Android.mk will deprecated someday.

PR #158, opened 2021-02-27
Fix crowbar smack think time

FreeSlave:
The time for think should be a global time not a UTIL_WeaponTimeBase.
Issue in Valve repo: https://github.com/ValveSoftware/halflife/issues/2296
Apparently it was actually updated in Steam version of Half-Life.


nekonomicon:
Yep. That's right.

PR #170, opened 2021-06-04
Haiku OS support

exstrim401:


a1batross:
Can you also add naming support for Haiku for CMake?

https://github.com/FWGS/hlsdk-xash3d/blob/master/cmake/LibraryNaming.cmake

Issue #165, opened 2021-04-04
[Big lolly] Wrong handgrenade ammo values.


malortie:
In the original biglolly release, picking up a handgrenade gives 2 ammo. The maximum carry value appears to be 6.

The following lines:

https://github.com/FWGS/hlsdk-xash3d/blob/cb1b9c6233ca762d4771b40f2a6534f750eb8add/dlls/weapons.h#L113

https://github.com/FWGS/hlsdk-xash3d/blob/cb1b9c6233ca762d4771b40f2a6534f750eb8add/dlls/weapons.h#L150

would need to be changed to this:

```cpp
#define HANDGRENADE_MAX_CARRY       6
#define HANDGRENADE_DEFAULT_GIVE    2
```

nekonomicon:
Thanks. Fixed.

Issue #164, opened 2021-04-04
[Big lolly] Wrong snark ammo values.


malortie:
In the original biglolly release, picking up a snark gives 3 ammo. The maximum carry value appears to be 12.

The following lines:

https://github.com/FWGS/hlsdk-xash3d/blob/cb1b9c6233ca762d4771b40f2a6534f750eb8add/dlls/weapons.h#L116

https://github.com/FWGS/hlsdk-xash3d/blob/cb1b9c6233ca762d4771b40f2a6534f750eb8add/dlls/weapons.h#L153

would need to be changed to this:

```cpp
#define SNARK_MAX_CARRY       12
#define SNARK_DEFAULT_GIVE    3
```

nekonomicon:
Thanks. Fixed.

Issue #163, opened 2021-04-04
[Big lolly] Wrong glock ammo values.


malortie:
In the original biglolly release, picking up a glock gives 30 ammo. The glock clip size appears to be 15.

The following lines:

https://github.com/FWGS/hlsdk-xash3d/blob/cb1b9c6233ca762d4771b40f2a6534f750eb8add/dlls/weapons.h#L124

https://github.com/FWGS/hlsdk-xash3d/blob/cb1b9c6233ca762d4771b40f2a6534f750eb8add/dlls/weapons.h#L140

would need to be changed to this:

```cpp
#define GLOCK_MAX_CLIP        15
#define GLOCK_DEFAULT_GIVE    30
```

PR #171, opened 2021-06-04
Fix weapon values in Big lolly

MrSwedish:
fix the pickup values according to #163 , #164 and #165


nekonomicon:
Thanks.

MrSwedish:
no problem

PR #176, opened 2021-06-07
Fix Goldsource build and add building with goldsource input into a workflow

FreeSlave:


mittorn:
conf.define in waf defining macros to 1 or 0, not empty string, so why you reverting back to ifdef?

nekonomicon:
This macros defined in code, but may be more clean to define this macros as 1.

FreeSlave:
The only Goldsource-related define configured from the outside is `GOLDSOURCE_SUPPORT`. And it was left as `#if defined`. Can be changed if necessary. `USE_VGUI_FOR_GOLDSOURCE_SUPPORT`, `SUPPORT_GOLDSOURCE_INPUT` and `USE_SDL2` are all defined depending on `GOLDSOURCE_SUPPORT` in the code.

PR #178, opened 2021-06-17
Android project

Velaron:
Probably requires more features, needs feedback.


Velaron:
Probably requires more features, needs feedback.

cl_dll/cl_util.h
```diff
@@ -108,7 +108,8 @@ inline int TextMessageDrawChar( int x, int y, int number, int r, int g, int b )
 inline int DrawConsoleString( int x, int y, const char *string )
 {
 	if( hud_textmode->value == 1 )
-		return gHUD.DrawHudString( x, y, 9999, (char*)string, 255 * g_hud_text_color[0], 255 * g_hud_text_color[1], 255 * g_hud_text_color[2] );
+		return gHUD.DrawHudString( x, y, 9999, (char*)string, (int)( (float)g_hud_text_color[0] * 255.0f ),
```
nekonomicon, line 111: It would be more clean to send separate PR.

nekonomicon:
Sorry, but I'm against gradle here.
There no reason to add yet another build system.
See discussion:
https://github.com/FWGS/hlsdk-xash3d/issues/105

Also, I'm not sure if it would be clean to have android project here.
I just think it may be bad idea to upload 20+ apks to releases in this repository.
May be It would be better to attach [mod-sample](https://github.com/FWGS/mod-sample) as submodule here specially for modders?

a1batross:
@nekonomicon 
I see no problem in gradle and even having Android project lying just there. I even wanted to do this like years ago.

* Android Studio natively supports it (well, it's the only buildsystem supported by Android Studio). 
* Most Android devs find it convenient and there is a lot of tutorials on how to build software with Android Studio already.
* It's possible to set it up to build native code with CMake, as opposed to deprecated Android.mk.

And I'm sorry too but `mod-sample` stinks. Nobody needs this unreadable mess. Yes, it can be built with lowest API level 1 but for what?

Most of the third-party mod devs just took existing Android project and used that, like one from CS16Client or one of yours. Nobody worked with `mod-sample` as far as I know.

PR #184, opened 2021-06-17
Comment unused variables

Velaron:


dlls/client.cpp
```diff
@@ -271,7 +271,7 @@ int Q_UTF8ToUChar32( const char *pUTF8_, unsigned int &uValueOut, bool &bErrorOu
 	bErrorOut = true;
 	return nBytes;
 
-decodeFinishedMaybeCESU8:
+//decodeFinishedMaybeCESU8:
```
nekonomicon, line 274: May be comment all block then?
Or place under `#if 0`?

Velaron: yeah didn't notice it's pretty much dead code

PR #192, opened 2021-10-16
add premake5 build script

iZarif:
I understand that the SDK can be built with CMake or Waf, but the more options the better, right? The script contains only two build configurations (Debug and Release), tested only on Linux and the build with LTO does not work correctly (despite the presence of the --lto command line option).


nekonomicon:
> but the more options the better, right?
Nope.
Who will update file lists and build options for all existing branches?

https://github.com/FWGS/hlsdk-xash3d/issues/105

a1batross:
Dude, I understand what you wanna say but we decided to not spawn another build system because we supported like seven different build scripts in old engine and it was a big headache.

HLSDK is more of exception because existing modders who use this as a base used CMake so it's still there.

If you wanna keep it for yourself, put it in contrib/iZarif directory so it'll not confuse random strangers who would like to use premake but something is broken, so they'll attach a bug but nobody except you is able to fix it.

PR #194, opened 2021-10-22
Get VGUI back (optionally)

FreeSlave:
I've brought back VGUI as an optional build flag and partially tested it on Linux Goldsource. MOTD and Scoreboard work.

Things to do:
- [x] Remove some code duplication happened during the code rewrite (some global variables have double declaration in VGUI and non-VGUI variant).
- [x] Check scoreboard in team deathmatch.
- [x] Check if bringing up the scoreboard on intermission works as expected.
- [x] Test menu.
- [x] Test spectator mode.
- [x] Update CMakeLists.txt so it supports Windows too.
- [x] Make it possible to keep scoreboard and motd non-vgui when vgui is enabled.
- [ ] Test building on macOS.
- [x] Test with Xash3D.
- [x] Add a [vgui-dev](https://github.com/FWGS/vgui-dev) submodule.
- [x] Add building with VGUI enabled to CI.
- [x] Remove servers browser.
- [ ] Test with nagist VGUI implementation?

### DIFF
--- a/cl_dll/vgui_TeamFortressViewport.cpp
+++ b/cl_dll/vgui_TeamFortressViewport.cpp
@@ -802,7 +802,7 @@ int TeamFortressViewport::CreateCommandMenu( const char *menuFile, int direction
 
 				// Get the button bound key
 				strncpy( cBoundKey, token, 31 );
-				cText[31] = '\0';
+				cBoundKey[31] = '\0';
 
 				// Get the button text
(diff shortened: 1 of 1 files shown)


FreeSlave:
I've implemented the option to build with vgui, but keeping non-vgui scoreboard and motd. May be it can be done dynamically in future, I just don't see much value in it right now.

If I add vgui-dev as a submodule, in which path it shuld be placed? In the repo root or utils/ directory?

PR #196, opened 2021-10-26
premake5: add arch suffix to library names

iZarif:


nekonomicon:
Is it possible to check `$CC -v`, `$CC -dumpmachine`, `$ARCH` or `$TARGET`?
Or even ask user about `$ARCH`?

iZarif:
> Is it possible to check `$CC -v`, `$CC -dumpmachine`, `$ARCH` or `$TARGET`? Or even ask user about `$ARCH`?

`$CC` is not always set, and if it is, it may not match the compiler that premake will use. Will `-dumpmachine` work better than `uname`? Asking the user about `$ARCH` is definitely a good idea, I was going to do it.

mittorn:
uname will show host architecture, dumpmachine shows target.
how do you want to compile without $CC?

iZarif:
> How do you want to compile without $CC?

In Premake there is no (or I have not found) way to get the path to the selected compiler, so there is no $CC

iZarif:
> Or even ask user about `$ARCH`?

Done, user can now specify the architecture using the --arch option

iZarif:
> That's incorrect.
> 
> You have to compile build.h, get all defined macros and select suffix according to engine/common/build.c.

Done, now if the user did not specify a compiler explicitly then gcc is implied. If the user does not have gcc installed the user will have to specify the architecture manually using the --arch option

FreeSlave:
Honestly I don't understand why we need any "contrib" stuff in this repo. Those who want alternative ways to build hlsdk can just keep it in their forks and merge/rebase upon master when needed.

iZarif:
> Honestly I don't understand why we need any "contrib" stuff in this repo. Those who want alternative ways to build hlsdk can just keep it in their forks and merge/rebase upon upstream master when needed.

I was just offering, I don't force anyone to merge it

PR #212, opened 2021-11-27
Remove unused source files

malortie:
See ValveSoftware/halflife#2996.


nekonomicon:
mpstubb.cpp has stubs for mp-only mods.

PR #221, opened 2021-12-08
Fixed wrong base class specified in CMultisource save restore.

malortie:
See ValveSoftware/halflife#3193.


a1batross:
Wouldn't that break existing saves?

nekonomicon:
I think it didn't break offsets.

PR #226, opened 2021-12-09
Fixed wrong base class specified in CEnvSound save restore.

malortie:
See ValveSoftware/halflife#3203.


malortie:
I finished inspecting save data fields class types and `IMPLEMENT_SAVERESTORE` base class for all entities.

No other similar issue has been found.

PR #231, opened 2021-12-11
Refactor EV_FireGlock2.

malortie:
Due to recent changes in `EV_FireGlock1` and `EV_FireGlock2`, both functions now have identical implementation.

The changes call `EV_FireGlock1` from `EV_FireGlock2`.

Modders will still be able to do custom implementation if needed.

**Important:** Mod branches might have different behavior. Make sure to check before merging from master.

### DIFF
--- a/cl_dll/ev_hldm.cpp
+++ b/cl_dll/ev_hldm.cpp
@@ -458,7 +458,8 @@ void EV_HLDM_FireBullets( int idx, float *forward, float *right, float *up, int 
 //======================
 //	    GLOCK START
 //======================
-void EV_FireGlock1( event_args_t *args )
+// Shared Glock fire implementation for EV_FireGlock1 and EV_FireGlock2.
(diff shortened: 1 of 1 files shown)


nekonomicon:
May be will be better to create separate static function and wrap it?
```
static void some_func(event_args_t *args)
{
...
}

void EV_FireGlock1(event_args_t *args)
{
some_func(args);
}


void EV_FireGlock2(event_args_t *args)
{
some_func(args);
}
```

malortie:
Would this do ?

```cpp
// Shared Glock fire implementation for EV_FireGlock1 and EV_FireGlock2.
static void EV_FireGlock_Impl( event_args_t *args )
{
  ...
}

void EV_FireGlock1( event_args_t *args )
{
	EV_FireGlock_Impl( args );
}

void EV_FireGlock2( event_args_t *args )
{
	EV_FireGlock_Impl( args );
}
```

PR #235, opened 2021-12-24
Simplify bullets effects code in ev_hldm.cpp

malortie:
See ValveSoftware/halflife#3212.


nekonomicon:
All good, but any compiler will optimize this code and ignore switch.

malortie:
Understood, but this is to simplify code.

I created separate PRs for the other changes, so that you may choose to merge or close.

nekonomicon:
> but this is to simplify code

I didn't see any reasons to remove empty switch-cases here.
It's already code style changes, not just refactoring.
It will be better to leave such code as is for changes in the future.

PR #238, opened 2021-12-24
Restore Hornetgun fire phase.

malortie:
`m_iFirePhase` should be saved so it can be correctly restored on game load.


a1batross:
But why?

malortie:
If `m_iFirePhase` is not saved, it will be reset to 0 on game load.

If a player performs a secondary attack after game load, this would ensure the cycle continues where it was.

a1batross:
Isn't that a weird edge case that player normally wouldn't hit?

malortie:
With `m_iFirePhase` unsaved, the behavior is reproductible:

1. Equip with the hornetgun
2. Fire 4 hornets in secondary mode (cycle is now halfway)
3. Save the game
4. Reload the game
5. Fire with the hornetgun in secondary mode

Expected: Hornetgun fire phase should continue from halfway to end cycle
Actual: Hornetgun fire phase restarts from the beginning

PR #241, opened 2022-01-18
Use filtered command buffer when string came from network

a1batross:
This is a client part of https://github.com/FWGS/xash3d-fwgs/pull/666

In theory, it must be compatible with the latest GoldSource, because the undocumented interface comes from it.

~~However, applying this patch will make client incompatible with older GoldSource and Xash3D builds, see how `Initialize( cl_enginefunc_t *, int )` is implemented.~~

Solved. We only use filtering if engine has cl_filterstuffcmd cvar.

### DIFF
--- a/cl_dll/cdll_int.cpp
+++ b/cl_dll/cdll_int.cpp
@@ -167,11 +167,16 @@ int DLLEXPORT Initialize( cl_enginefunc_t *pEnginefuncs, int iVersion )
 	if( iVersion != CLDLL_INTERFACE_VERSION )
 		return 0;
 
-	memcpy( &gEngfuncs, pEnginefuncs, sizeof(cl_enginefunc_t) );
+	// for now filterstuffcmd is last in the engine interface
+	memcpy( &gEngfuncs, pEnginefuncs, sizeof(cl_enginefunc_t) - sizeof( void * ) );
 
 	if( gEngfuncs.pfnGetCvarPointer( "cl_filterstuffcmd" ) != 0 )
 	{
 		gEngfuncs.pfnFilteredClientCmd = gEngfuncs.pfnClientCmd;
+	}
+	else
+	{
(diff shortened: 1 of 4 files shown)


FreeSlave:
I would prefer to keep the compatibility with older versions. Currently this hlsdk can work even with WON versions of Goldsource. Also it should continue to work with original Xash (non-fwgs).
Can we somehow check for this function in the runtime and use the old method as a fallback? Or if it's not possible, put it under the macro.

a1batross:
Unfortunately, during initialization, client.dll doesn't know the size of engine funcs passed to it and uses what it was compiled with. It allows interface extension from engine side but doesn't from client. This is also problem for both Render and Mobility interfaces too.

Can't think of better solution here. We may deprecate old mobility interface since I don't want to maintain compatibility with old engine here...

a1batross:
But I think we shouldn't leave this unpatched. Executing commands from server side is a known problem for years.

a1batross:
Ah, I get it now!

GoldSource and Xash with support for command filtering expose "cl_filterstuffcmd" cvar. If we can get a pointer to it, high chances that engine supports this extended interface.

mittorn:
It still reads engfuncs out of bound in memcpy. Maybe we need some way to determine array size based on engine version

a1batross:
> It still reads engfuncs out of bound in memcpy. Maybe we need some way to determine array size based on engine version

Yes I thought about it... oh well.

Maybe the same check can determine the size of the struct.

a1batross:
@nekonomicon @mittorn 

do you have any comments? I updated this branch.

Probably not in the best way, it will break if we add something to the interface again.

a1batross:
сс @FreeSlave

nekonomicon:
LGTM.

FreeSlave (approved):

nekonomicon (approved):

FreeSlave:
Should we merge it already?

PR #247, opened 2022-04-09
Use valid animation extension for pipe.

malortie:
This issue also appears in the original mod.

The player model does not have animations suffixed pipe, so it defaults to the first sequence (idle).

The following changes make the pipe use 'crowbar' anim extension so it can pick a valid sequence.


malortie:
@nekonomicon Can you please check the cancelled build, and in PR #248 ?

nekonomicon:
It's just deprecated Visual Studio version, nothing critical.

PR #250, opened 2022-04-21
Detect AMD64 target in CMake

a1batross:
cc @nekonomicon 

cc @FreeSlave I probably did something wrong for MSVC, so pls check it too!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,11 @@ set(SERVER_LIBRARY_NAME "hl" CACHE STRING "Library name for Linux/MacOS/Windows"
 # MAIN BUILD CODE \
 ###################\
 
-if(CMAKE_SIZEOF_VOID_P EQUAL 8 AND NOT 64BIT)
+# Only append -m32 on x86 targetting 64-bit mode, when not cross compiling
+if(NOT 64BIT AND
+	NOT CMAKE_CROSSCOMPILING AND
+	CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" AND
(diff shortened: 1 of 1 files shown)


FreeSlave:
Well, the produced libraries work, so looks good to me

nekonomicon:
@FreeSlave what about 32-bit target on 64-bit windows?

FreeSlave:
I think all github actions hosts are 64-bit.
The libraries produced by our CI are all 32-bit.

FreeSlave:
Okay, I've tried building on this branch, without adding a parameter and amd64 libraries were generated.
If I do the same on master, cmake just stops at the line `error("UNDONE: set 32 build flags")` because of the error `Unknown CMake command "error".`

nekonomicon:
@a1batross It works incorrectly.
Need to add following strings:
`CMAKE_SYSTEM_PROCESSOR STREQUAL "x64"` for Windows(May be something yet)
`CMAKE_SYSTEM_PROCESSOR STREQUAL "amd64"` for BSD

See:
https://stackoverflow.com/questions/70475665/what-are-the-possible-values-of-cmake-system-processor
https://stackoverflow.com/questions/45125516/possible-values-for-uname-m/45125525#45125525

Also, library naming scheme didn't work here.

nekonomicon:
Or will be better to never check `CMAKE_SYSTEM_PROCESSOR`.

nekonomicon:
Also, for MSVC probably need to set `CMAKE_EXE_LINKER_FLAGS "/MACHINE:x86"`

nekonomicon:
Does not actually.

PR #251, opened 2022-05-07
Apply CROWBAR_FIX_RAPID_CROWBAR to knife

mmillar-bolis:
Greetings and thank you for your time and consideration!

I observed that OpFor's knife still exhibits the rabid melee bug, even when building the server with the `CROWBAR_FIX_RAPID_CROWBAR` flag. This correction applies only to the crowbar, but the same code can fix the knife.

I don't know whether you would want to incorporate this code into knife.cpp as it is, or change the variable name and then apply it here.

Thank you again!


FreeSlave:
Looks good.
@nekonomicon can you merge it?

PR #264, opened 2022-07-09
Build with GoldSource input by default on Windows and Linux

FreeSlave:
Note that I didn't touch building with waf. It still requires `--enable-goldsrc-support` option. I postponed it before we decide on what to do with waf as currently I see cmake is the main way to build the sdk. Maybe waf is gonna be used only for Android builds.

I also removed building with xash3d-fwgs input only from the github workflow as I don't see its value.

Next step will be to change the mention of GOLDSOURCE_SUPPORT option in build instructions (I'll do it in another PR with general README.md update).

@nekonomicon can you check if I did it right? The idea is that `GOLDSOURCE_SUPPORT` should set by default only when building x86 targets on Windows and Linux.


nekonomicon:
Waf probably will be removed.
For android we need cmake and gradle.

a1batross:
I will update wscripts accordingly.

Issue #270, opened 2022-07-10
Proof of concept: advanced background music management


FreeSlave:
A lot of mods used fmod for playing music. We haven't yet incorporated fmod in any mod branch that originally used it. And these days we probably shouldn't do so. We should seek for a lightweight opensource alternative. @nekonomicon suggested we should try [miniaudio](https://github.com/mackron/miniaudio).

This has a potential of improving the music experience in mods, e.g. not interrupting the music on changelevel.

PR #273, opened 2022-07-11
HL:Invasion Add MiniAudio Music implementation

RoyShapiro:
Hello again!

I've added Miniaudio music implementation to Invasion mod, as promised.
It should now be much more portable.

The CMakeList.txt still uses null music player by default, so the music implementation is selectable at compile time.
I'm thinking of dropping both the GStreamer ("bloat" (c) @nekonomicon ) and the original FMod (proprietary and incredibly outdated) implementations.

What do you think?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,7 @@
 [submodule "vgui_support"]
 	path = vgui_support
 	url = https://github.com/FWGS/vgui_support
+
+[submodule "miniaudio"]
+	path = miniaudio
+	url = https://github.com/mackron/miniaudio
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ option(BUILD_CLIENT "Build client dll" ON)
 option(BUILD_SERVER "Build server dll" ON)
 option(GOLDSOURCE_SUPPORT "Build goldsource compatible client library" OFF)
 option(USE_GSTREAMER "Enable GStreamer." OFF)
+option(USE_MINIAUDIO "Enable Miniaudio." OFF)
 option(USE_FMOD "Enable FMOD." OFF)
 
 if (CMAKE_SIZEOF_VOID_P EQUAL 4 OR
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Things fixed while transferring code:
 - Sniper rifle zoom states didn't work, m_pPlayer->pev->fov needed to be equalized to m_pPlayer->m_iFOV.
 - VGUI used to hard-crash, partially fixed: CImageLables in vgui_OrdiControl that caused the crash were replaced with CommandButtons, skipTime added to keypad, OrdiMenu and OrdiControl in order to fix double mouse key presses, +1 added to radiomsg.cpp to fix the butts of the text messages, HUD health display didn't work due to Health Msg system receieving Battery-related data, fixed by manually re-coping a part of old code and paritioning it properly.
 - Also fixed .txt files and fonts not loading due to backslashes, VGUI folder name being written in capitals and such.
-- Music not playing fixed by creating a GStreamer implementation.
+- Music not playing fixed by creating Miniaudio (thanks @nekonomicon) and GStreamer implementations.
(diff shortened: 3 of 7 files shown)


a1batross:
Should it be on server though? For Invasion it's ok since it's a
singleplayer mod but for generic implementation it wouldn't work nicely in
multiplayer.

пн, 11 июл. 2022 г., 23:51 nekonomicon ***@***.***>:

nekonomicon:
Unfortunately, original invasion uses fmod on server side.
I think it must be disabled for multiplayer.

RoyShapiro:
@nekonomicon 
Can do. I'd still leave the "null" option included for systems that may not support miniaudio (VGUI still remains a portability issue).
Invasion is a purely singleplayer mod, if you don't count the photoshopped-out button in btns_main.bmp :D

@a1batross
This particular code is non-generic, it comes with an Invasion-specific implementation of "`trigger_music`" that uses playlists. I'm not aware if this was ever used word for word in any other mods. That said, it seems like Poke646 and Vendetta, off the top of my head, also use server-side FMod implementation. So, this code may be useful there as well, but will likely need adaptation. I don't know the details, though.

nekonomicon:
Poke646 reads playlist and uses FMod on client side.

RoyShapiro:
@nekonomicon 
Okay, good to know.

I've removed gstreamer / fmod / null implementations and options in a new commit, https://github.com/RoyShapiro/hlsdk-xash3d/commit/418ae2d41d8ee4a402f597c0db038bc6bab246a2.
I've also added the option -DDISABLE_MINIAUDIO instead to disable the current one if necessary. It uses defines within music.h and music.cpp themselves to disable miniaudio code, this is much neater.

I've also tried to find a way to disable this in multiplayer,
see music.cpp in commit https://github.com/RoyShapiro/hlsdk-xash3d/commit/418ae2d41d8ee4a402f597c0db038bc6bab246a2 lines 413 and 441 respectively (_I may be doing something wrong there, but I saw other triggers using the same code for that purpose_), but when I tried to test that, by forcing the game into multiplayer, I got `Error: No spawn function for trigger_music` and it didn't spawn anyway. IDK why, it most definitely has a spawn function and work fine in singleplayer. Weird O.o.

nekonomicon:
Try to use [UTIL_Remove](https://github.com/FWGS/hlsdk-portable/blob/master/dlls/util.cpp#L1501=).
It will prevent spawn.

RoyShapiro:
@nekonomicon 
Okay, it's probably going to be a really stupid question, but where would I use UTIL_Remove? AFAIK the first method called would be "Spawn", but in multiplayer, it says that it doesn't have spawn, even though it does. O.o

Anyway, I tried moving the music player code **to the client** instead as suggested, and was **successful**.
Haven't pushed anything, yet. Should I? Or will we leave it on the server?

The code turned out to a little less... Compact.
There a small class in `hud.h \ hud.cpp` now, that reads messages coming from server.
Also, there is an `int` variable on the server side, that's used for `REG_USER_MSG`.
Most of those are defined in `player.cpp`. I temporarily moved it to my own music.cpp.

I guess it stands to reason to move it to `player.cpp` where the rest of them reside, does it?

a1batross:
There is multiplayer part of Invasion? I never played that mod to be honest. :)

RoyShapiro:
@a1batross 
No, there is not :D 
But since it feels like we want this code to be more generic and applicable to other mods, potentially, before merging, I figured I had to test it that way. So I kinda forced it upon the mod by temporarily editing gameinfo.txt and liblist.gam files.

a1batross:
Under generic support, I've meant https://github.com/FWGS/hlsdk-portable/issues/270

RoyShapiro:
@a1batross Oh, I see, so @FreeSlave wanted to tackle the issue as well. I didn't know that, so my attempts to kinda do the same are unrelated and I wasn't initially planning to do a generic implementation. I just wanted to "debloat" my previous code mess :D

Anyway, Invasion has some peculiarities, like the playlist format.

I've just pushed the client side version here, https://github.com/FWGS/hlsdk-portable/pull/273/commits/3e62e6b719f8bcae8732d1de1c8476b28c054249
it's still pretty Invasion-specific, but the same idea, sans the playlist format, can be relatively easily applicable elsewhere.

nekonomicon:
LGTM.
We can try to fix windows build later.

PR #280, opened 2022-07-18
Add WEAPONS_ANIMATION_TIMES_FIX macros. Fix #172

### LINKED ISSUE #172: Use Fixes from ValveSoftware/halflife/issues/2495
https://github.com/ValveSoftware/halflife/issues/2495
includes nice animation fixes also in code!
look at it!

FreeSlave:
@nekonomicon we can apply the fix for crowbar idle animation even without this macro as crowbar plays idle animation only when another macro is enabled.
Also we can remove the line `m_flTimeWeaponIdle = UTIL_WeaponTimeBase() + UTIL_SharedRandomFloat( m_pPlayer->random_seed, 10.0f, 15.0f );` from satchel's `Deploy` altogether as `m_flTimeWeaponIdle` is rewritten by a subsequent call to `DefaultDeploy` anyway.

What do you think?


nekonomicon:
I think we can do It.

FreeSlave:
@nekonomicon added commits

PR #286, opened 2022-07-19
Rename false_vgui to fake_vgui

FreeSlave:
For the sake of correctness


a1batross:
Isn't fake_vgui can use actual vgui headers?

a1batross (approved):

FreeSlave:
These are stub implementations. May be they can be moved to .cpp and then headers from vgui can be used.

PR #306, opened 2022-08-06
Replace usage of m_ForceSendAnimations with entvars

FreeSlave:
To avoid changing the class data layout


a1batross:
As far as I remember, oldbuttons is used in prediction code. Are you sure it won't collide with it?

FreeSlave:
I think it's used only for player entity.

Here I use `oldbuttons` of weapon entity.

PR #322, opened 2022-08-29
Add F function

FreeSlave:
Tested HUD_ChatInputPosition with VGUI and without. Works as expected.

Do we need this in master?

### DIFF
--- a/cl_dll/Exports.h
+++ b/cl_dll/Exports.h
@@ -1,0 +1,66 @@
+#include "exportdef.h"
+#include "mobility_int.h"
+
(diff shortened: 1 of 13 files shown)


FreeSlave:
Okay, I've found out why `HUD_ChatInputPosition` didn't work before. It's because we have the definition in `vgui_SpectatorPanel.cpp`, but miss a declaration with extern "C".

So I guess there's no real reason for now to use `F`. I'll do a separate PR to fix `HUD_ChatInputPosition`.

PR #328, opened 2022-09-18
server: func_tank: prevent domain error on barrel adjust

a1batross:
When target is too close to the tank origin, `d2 - r2` expression may become negative causing domain error on square root, and poisoning other fields and even other entities with NaN

This can be easily tested by jumping on a tank head on `c2a5b`.

I'm not sure if we should take absolute value or zero in this case.

### DIFF
--- a/dlls/func_tank.cpp
+++ b/dlls/func_tank.cpp
@@ -626,12 +626,14 @@ void CFuncTank::AdjustAnglesForBarrel( Vector &angles, float distance )
 		if( m_barrelPos.y )
 		{
 			r2 = m_barrelPos.y * m_barrelPos.y;
-			angles.y += ( 180.0f / M_PI_F ) * atan2( m_barrelPos.y, sqrt( d2 - r2 ) );
+			if( d2 > r2 )
+				angles.y += ( 180.0f / M_PI_F ) * atan2( m_barrelPos.y, sqrt( d2 - r2 ) );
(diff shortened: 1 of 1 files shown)


a1batross:
Nah, this is not correct.

With `fabs()` tank head begins to turning by few degress, like it's in panic. :)

Will update this branch with another solution

a1batross:
Feels a bit more stable with just doing nothing if target is too close. It's still twitching but less.

FreeSlave:
Funny, you can be teleported to the world origin if standing on a certain spot on top of Bradley on `c2a5b`.
It doesn't happen with this fix.

FreeSlave (approved):

nekonomicon:
What if just do nothing for negative value?

a1batross:
@nekonomicon yeah, in last version I don't change any angles.

nekonomicon:
Where?
I see fabs.

nekonomicon:
@a1batross all good?
https://github.com/FWGS/hlsdk-portable/compare/35cd765a775204bf8bbb68b292e0133d20589306..17281f80af3522318f43c8d83ed404aee3fd8d73

a1batross:
@nekonomicon no, it's not... I force-pushed `fabs`-less changes. But haven't verified them, and these were old changes. O_o

Well, should be OK now.

PR #344, opened 2022-11-23
Don't print 'Node Graph out of Date' messages in non-developer mode

FreeSlave:
Note that in Valve repo the message printing is commented out https://github.com/ValveSoftware/halflife/blob/master/dlls/nodes.cpp#L1490
As a compromise we can still print it, but only in developer mode.

### DIFF
--- a/dlls/nodes.cpp
+++ b/dlls/nodes.cpp
@@ -24,6 +24,7 @@
 #include	"nodes_compat.h"
 #include	"animation.h"
 #include	"doors.h"
+#include	"game.h"
 
 #define	HULL_STEP_SIZE 16// how far the test hull moves on each step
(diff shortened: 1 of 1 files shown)


a1batross:
I think we forced printing it because when nodes wasn't shipped with the game for some reason, nodes generation without this iconic message make the game look like it's frozen.

FreeSlave:
I think the message stays for too long, longer than the node graph building usually takes. And there's no way to control the duration. But it's up to you.

a1batross:
Player knows when nodes generation is finished :)

nekonomicon:
> I think the message stays for too long, longer than the node graph building usually takes.

Platform/CPU/Optflags-specific.

PR #355, opened 2023-02-19
Speedup cmake configuration on Windows

FreeSlave:
Due to a big number of platform checks the initial cmake configuration on Windows is pretty slow. It mostly affects automated builds as it always starts as an initial build in the container.

The goal of this PR is to increase the awareness on the problem and maybe spark a discussion on possible solutions. This might be not suitable for merging into master branch.


a1batross:
CMake should support either querying list of defined symbols, or defined value, or finally have multithreaded configure. 

Maybe precomipiling build.h may help? Btw newer engine version is even more complex and importing it can make CMake checks even more slower.

a1batross:
Few other ideas:

* Newer engine revision of this header just sets the XASH_PLATFORM, XASH_ARCHITECTURE values. Sadly, just querying the value isn't possible without running a compiled binary. But this potentially can be exploited when we're not crosscompiling. Anyway, the only build system capable of easy crosscompiling in our HLSDK is waf. =/

* The checks in this file can be reordered in such a way to check commonly used targets like win32-i386 and win32-amd64 first and don't execute next if these are matched.

a1batross:
As for compiling on the target machine, you can just compile Q_buildarch and Q_buildos functions from the engine, these are completely standalone and then parse the output, it also was made to be easily parseable with simple regex.

a1batross:
See the alternative solution in https://github.com/FWGS/hlsdk-portable/pull/368

FreeSlave:
Closing as it got resolved in https://github.com/FWGS/hlsdk-portable/pull/368

Issue #263, opened 2022-07-09
Making this repo more accessible and acknowledged


FreeSlave:
Recently we had a discussion on FWGS discord servers about how we make this repo more popular among mod makers.
While this repo contains source codes for a lot of mods that might be useful to mod-makers, it's still not well-known among common crowd and mostly used by Linux/Xash3D enthusiasts.

I propose the following steps:

- Change the name. `hlsdk-xash3d` sounds like it's exclusively for xash3d. It's not true for long since the time GoldSource input support was added to the client and VGUI has got returned. @nekonomicon suggested it should be called `hlsdk-portable`. I like this as it emphasizes on the strength of this hlsdk version.
- The README.md and the repo description must be clear about the goal of this hlsdk (be portable across engines and platforms and provide mods implementations).
- Build with GoldSource support by default. Minus one explicit action by a user in most cases. The only downside is that you must have SDL2 dev headers installed on Linux. Also those who are not interested in GoldSource input support, will need to turn it off themselves.
- Provide better build instructions. Android build instructions are missing, Windows ones are outdated. There are many mod-makers that are not programmers and they just want to change a small thing, like a HUD color, which doesn't require programmer's skills, so the main problem for them would be building. We must decide what is the best way for a typical user to build libraries (via cmake or waf) and write more detailed instructions accordingly (with direct links to cmake/python websites). Keep in mind that the majority of users are Windows users. Details about `msbuild` must be provided too. Probably instuctions should be moved to the Wiki as README.md is gonna be too bloated with all these details.
- Maybe we should provide Visual Studio projects, but considering there are different VS versions, it's probably better to stick to cmake. It has GUI on Windows and becomes more acknowledged among Windows users. Also adding another build system is going to lead to more maintanence costs.

How this repo should be advertised (a *selling* point):

- Portable. Among engines (GoldSource and Xash3D, both original Xash3D one and Xash3D FWGS) and among platforms.
- A lot of mods re-implemented. The universal codebase as each mod is a branch.
- Ready-to-use automated builds for mods (including user's mods if they're gonna fork the repo).
- A lot of fixed bugs. Probably we need make a list and place it somewhere, e.g. on repo Wiki page.

Probably we should make it more clear that automated builds won't work on Windows-XP. I can add a manual workflow that would allow building with msvc6. That means however that we have to support `compile.bat` as a build-system. Or come up with something better.

[...]

PR #359, opened 2023-03-09
PlayStation Vita support

fgsfdsfgs:
Adds PSVita support to both waf and cmake scripts. Complementary to FWGS/xash3d-fwgs#1251.
Building the libraries for `psvita` requires the same dependencies as building the engine for it.

**Building with waf:**
```
./waf configure -T release --psvita
./waf build
```

**Building with CMake:**
```
mkdir build && cd build
cmake -G"Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE="$VITASDK/share/vita.toolchain.cmake" -DCMAKE_PROJECT_HLSDK-PORTABLE_INCLUDE="$VITASDK/share/vrtld_shim.cmake" ..
make -j
```


scripts/waifulib/xcompile.py
```diff
@@ -434,6 +434,75 @@ def ldflags(self):
...
+	def cflags(self, cxx = False):
+		cflags = []
+		# arch flags
+		cflags += ['-D__vita__', '-mtune=cortex-a9', '-mfpu=neon']
+		# necessary linker flags
+		cflags += ['-Wl,-q', '-Wl,-z,nocopyreloc']
+		# this optimization is broken in vitasdk
+		cflags += ['-fno-optimize-sibling-calls']
+		# disable some ARM bullshit
+		cflags += ['-fsigned-char', '-fno-short-enums', '-Wno-attributes']
```
a1batross, line 483: We should be able to build without `-fsigned-char` now.

But `-fno-short-enums` should be necessary.

scripts/waifulib/xcompile.py
```diff
@@ -434,6 +434,75 @@ def ldflags(self):
...
+		# base include dir
+		cflags += ['-isystem %s/arm-vita-eabi/include' % self.vitasdk_dir]
+		# SDL include dir
+		cflags += ['-I%s/arm-vita-eabi/include/SDL2' % self.vitasdk_dir]
+		# the game wants GNU extensions
+		if cxx:
+			cflags += ['-std=gnu++17', '-D_GNU_SOURCE']
```
a1batross, line 490: Don't we enable _GNU_SOURCE only when we actually use it?

scripts/waifulib/xcompile.py
```diff
@@ -434,6 +434,75 @@ def ldflags(self):
...
+		# necessary linker flags
+		cflags += ['-Wl,-q', '-Wl,-z,nocopyreloc']
+		# this optimization is broken in vitasdk
+		cflags += ['-fno-optimize-sibling-calls']
+		# disable some ARM bullshit
+		cflags += ['-fsigned-char', '-fno-short-enums', '-Wno-attributes']
+		# base include dir
+		cflags += ['-isystem %s/arm-vita-eabi/include' % self.vitasdk_dir]
+		# SDL include dir
+		cflags += ['-I%s/arm-vita-eabi/include/SDL2' % self.vitasdk_dir]
```
a1batross, line 487: SDL2 shouldn't be added globally here and instead be in uselib.

But this is a minor issue, I wouldn't change it right now but refactor it in engine later. Maybe I would do that myself.

scripts/waifulib/xcompile.py
fgsfdsfgs: `-fsigned-char` is actually the default in vitasdk gcc now, so removed.

scripts/waifulib/xcompile.py
fgsfdsfgs: Yeah, I don't remember why I put that in. Removed.

nekonomicon:
@fgsfdsfgs please add build instructions to readme.md.

a1batross:
@nekonomicon do you really wanna pile up all build instructions in one single readme md file?

nekonomicon:
@a1batross [issue for any suggestions](https://github.com/FWGS/hlsdk-portable/issues/263)

fgsfdsfgs:
> @fgsfdsfgs please add build instructions to readme.md.

Done.

PR #362, opened 2023-04-06
Fix of shitcode in CSatchel::Deploy (satchel.cpp)

xorusr:


a1batross:
Mind your language.

But actually, [this](https://github.com/FWGS/hlsdk-portable/commit/058626fc18d76fa1e33685c22a15993b68ef79a3) change breaks prediction with original DLLs. 
cc @FreeSlave 

See how it calls shared random in [original HLSDK](https://github.com/ValveSoftware/halflife/blob/master/dlls/satchel.cpp#L292), the calls to shared random is supposed to be equal between client and server weapons code.

PR #368, opened 2023-04-17
Buildsystems overhaul

a1batross:
* LibraryNaming.cmake got a speedup
* Wscripts were reworked to address easier maintaining
* There is now a common mod configuration file, called mod_options.txt.

It is fairly simple "key=value # comment" format, all boolean keys (detected by values "ON" or "OFF"), are automatically declared as options in CMake and macros in the code. Strings are only imporant to buildsystem, like directories where to put DLLs.


FreeSlave:
I checked this on Windows, the cmake configuration is 5 seconds on this branch against 17 seconds on master. Good stuff.

a1batross:
@FreeSlave what do you think about mod_options?

FreeSlave:
> @FreeSlave what do you think about mod_options?

idk, it doesn't bother me. The options are still handled by gui configurators (like cmake-gui), so it's okay.
But it seems that changing the mod_options.txt file and reconfiguring the project with cmake on the same build directory doesn't affect the options: their values stayed configured the way it was on the first configuration. This can be misleading. But I think it's the same as if you change the default value in CMakeLists.txt.

FreeSlave (approved):

nekonomicon (approved):

PR #369, opened 2023-04-23
Fix hgrunts dropping weapons again. Fix magical constants in hgrunt code.

FreeSlave:


stephen-cusi:
This should be merged into the fix branch

FreeSlave:
> This should be merged into the fix branch

Why so? Does it break anything besides the exploit?

stephen-cusi:
> > This should be merged into the fix branch
> 
> You mean that it doesn't belong to master? Why so? Does it break anything besides the exploit?

Sorry, I was careless

PR #377, opened 2023-05-30
cmake: LibraryNaming: fix Android check

Velaron:
Would also be great to merge this into mobile_hacks branch after approval


a1batross:
Does waf have same mistake?

Velaron:
I use existing CMakeLists for Android build, but probably waf is not affected by this problem (engine/mainui builds fine)

a1batross:
Yep, waf isn't affected, as it wasn't changed and didn't require a speedup optimizations.

PR #379, opened 2023-06-12
ARMv8 neon support for studio util

MoeMod:


MoeMod:
> Does it make any difference?
> 
> Of course, compared with release build.

I found that AngleQuaternion and QuaternionSlerp takes 30%+ cpu time in multiplayer CS, so in such scenario it may save some performance.

a1batross:
It would make sense to add such changes to engine's libpublic as well. And
write tests, so the results are accurate. :).

nekonomicon:
BTW may be use [SSE version](http://gruntthepeon.free.fr/ssemath/sse_mathfun.h)(useful for x86 and e2k) + [SIMD everywhere](https://github.com/simd-everywhere/simde)?
It can give performance boost on many architectures.

a1batross:
@nekonomicon it would be nice to at least start with ARM.

a1batross (approved):
I mostly approve. Maybe it only needs slight reformatting to be in line with the other SDK code.

Adding a separate SSE variant is out of the scope of this PR.

PR #389, opened 2023-09-09
WIP: Add mod option for legacy mods that had _i386 postfix on Linux builds

a1batross:
On 32-bit x86 Linux turns server library to be `hl_i386.so` instead of `hl.so`.

Probably incompatible with old WON HLDS builds, but the purpose is to be compatible with legacy naming in `liblist.gam`.


a1batross:
Modern GoldSource automatically strips everything after '_', i.e. even if liblist.gam has `hl_i386.so`, engine will load `hl.so` instead. 

I don't think it makes sense to target old GoldSrc builds.

PR #412, opened 2023-12-21
HL25 HUD

a1batross:
This set of patches adds support for available since HL25 update HD set of sprite sheets.

Because not every player might like the new HUD look, there is a new cvar `hud_allow_hd` that limits the usage of HD sprites and reverts some style changes, like battery HUD being super close to the left and HUD numbers are shifted down a bit to match the icons (additionally, it fixes HUD inconsistency bug caused by this change, see ValveSoftware/halflife#3728).

There is no way to switch between old and new HUD by toggling `hud_allow_hd` cvar without restarting a level, so I wonder if we should use Xash3D's FCVAR_LATCH extended flag to notify the player that the change will be skipped until server restart.

### DIFF
--- a/cl_dll/ammo.cpp
+++ b/cl_dll/ammo.cpp
@@ -873,6 +873,7 @@ int CHudAmmo::Draw( float flTime )
 
 	// Does this weapon have a clip?
 	y = ScreenHeight - gHUD.m_iFontHeight - gHUD.m_iFontHeight / 2;
+	y += gHUD.m_iHudNumbersYOffset; // a1ba: fix HL25 HUD vertical inconsistensy
 
 	// Does weapon have any ammo at all?
 	if( m_pWeapon->iAmmoType > 0 )
--- a/cl_dll/ammo.cpp
+++ b/cl_dll/ammo.cpp
@@ -75,10 +75,7 @@ void WeaponsResource::LoadWeaponSprites( WEAPON *pWeapon )
 {
 	int i, iRes;
 
-	if( ScreenWidth < 640 )
-		iRes = 320;
-	else
-		iRes = 640;
+	iRes = GetSpriteRes( ScreenWidth, ScreenHeight );
 
 	char sz[256];
 
--- a/cl_dll/ammo.cpp
+++ b/cl_dll/ammo.cpp
@@ -325,16 +325,19 @@ int CHudAmmo::VidInit( void )
 	// If we've already loaded weapons, let's get new sprites
 	gWR.LoadAllWeaponSprites();
 
-	if( ScreenWidth >= 640 )
-	{
-		giABWidth = 20;
-		giABHeight = 4;
-	}
+	const int res = GetSpriteRes( ScreenWidth, ScreenHeight );
+	int factor;
+	if( res >= 2560 )
+		factor = 4;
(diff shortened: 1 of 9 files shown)


a1batross:
I reverse-engineered the exact numbers (like switching between resolutions in `GetSpriteRes` and `iHudNumbersYOffset`) but I wrote the code the way I wanted, so the function I mentioned doesn't exist in decompiled client.so.

tmp64:
While the code matches HL, I think ammo bars should scale with HUD size for consistency

https://github.com/FWGS/hlsdk-portable/blob/e8db2fa3545670c3a0a48a5887215121fa7351ba/cl_dll/ammo.cpp#L328-L337

a1batross:
@tmp64 yeah, I checked that specifically in decompile, and they haven't touched that.
Dunno, it still looks good enough to me.

FreeSlave:
They also got the train sprite wrong as described in https://github.com/ValveSoftware/halflife/issues/3510

a1batross:
@FreeSlave about train code not referencing train sprites hud.txt: I don't think the behavior should be changed that much.

a1batross:
@tmp64 I fixed the ammo bar scaling you mentioned in last commit.

FreeSlave:
Imagine the following situation.
The mod uses this code and provides sprites for high resolutions. Some sprites are larger than 256x256 and the pre-anniversary goldsource wouldn't be able to load them. However the developer wants the mod to work on both current and legacy branches.
So we need to detect whether the client is running on the legacy version, and don't load high res sprites in this case and treat everything as if `hud_allow_hd` was set to 0.

a1batross:
True, that's planned to do.

nekonomicon:
Is it ready to merge?

a1batross:
I'm too lazy to check whether legacy GoldSrc can support this. 

Until we get confirmation that it works, or otherwise I make some detection to disable the HD HUD, I don't think it's ready to be merged.

a1batross:
Also, this:

```
There is no way to switch between old and new HUD by toggling hud_allow_hd cvar without restarting a level, so I wonder if we should use Xash3D's FCVAR_LATCH extended flag to notify the player that the change will be skipped until server restart.
```

I'm still not sure about this. Should we just use LATCH flag from Xash3D (and hope that it won't break anything for GoldSrc) or do some workaround?

tmp64:
BHL does this by checking if one of the new sprites exists  
https://github.com/tmp64/BugfixedHL-Rebased/blob/1a1f601b50217ef8ac5f1d84f7813361c6a36ae1/src/game/client/hud.cpp#L114-L120

FreeSlave:
It's not what I'm talking about.
You can't rely on checking for resources.
Imagine an author wants a mod to run on both steam_legacy and the current HL versions, using benefits of support for highres sprites when it's possible. The old engine doesn't support big sprites (I actually need to check what happens if it tries to do so).

nekonomicon:
>I'm too lazy to check whether legacy GoldSrc can support this.

>Until we get confirmation that it works, or otherwise I make some detection to disable the HD HUD, I don't think it's ready to be merged.

Ok.
Who really need legacy GoldSource support, probably can write version check by yourself.

PR #413, opened 2023-12-21
[WIP] Warn mod developer about reaching the GoldSrc sentences limits

a1batross:
Quick draft for an idea suggested by @FreeSlave 

Should it be more complicated? Should this check be removed without GOLDSOURCE_SUPPORT, which is only for client.dll now?

### DIFF
--- a/dlls/sound.cpp
+++ b/dlls/sound.cpp
@@ -1341,6 +1341,15 @@ void SENTENCEG_Init()
 
 	g_engfuncs.pfnFreeFile( pMemFile );
 
+	if( gcallsentences >= 2048 )
+	{
+		ALERT( at_warning, "NOTE: this mod might not work properly under GoldSource (post-anniversary update) engine: more than 2048 sentences\n" );
+	}
+	else if( gcallsentences >= 1536 )
+	{
+		ALERT( at_warning, "NOTE: this mod might not work properly under GoldSource (pre-anniversary update) engine: more than 1536 sentences\n" );
+	}
+
 	fSentencesInit = TRUE;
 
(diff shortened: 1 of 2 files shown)


FreeSlave:
Will you apply proposed changes and merge it? You can request review from other members.

nekonomicon:
What I can say, there may be useful dynamic arrays for platforms with low memory.
If you don't have other ideas or suggestions I will merge it.

FreeSlave:
Can 48 KB make a significant difference in this context though?
And if it's a real problem, are there any other places in hlsdk where we could optimize for size? Maybe add another compile-time variable then?

a1batross:
The sentence array size definitely must match the engine. There is no reason not to. If some port is limited by memory, the sizes must be decreased in engine first, and then applied to HLSDK.

As for dynamic allocations... it might be useful, I agree. But this isn't something I wanted to do in such PR. 

-----

By the way, why I didn't hide these warnings under SUPPORT_GOLDSOURCE_ENGINE? To me, it makes sense.

FreeSlave:
The `GOLDSOURCE_SUPPORT` is used only on client now. You'll need to forward the definition in server CMakeLists as well.

nekonomicon:
> Can 48 KB make a significant difference in this context though?

It's may be crititcal on psp and ps vita at least.
With current change will be 64KiB on disk and in memory.
In android builds will be 256KiB on disk.
Plus array on engine side eats the same free space too.

> And if it's a real problem, are there any other places in hlsdk where we could optimize for size?

Array for textures in pm_shared.c probably.

PR #424, opened 2024-01-13
Finding a reason of server crash on the online change

d47081:
After enabling dynamical add/remove bots on online change (by min/max settings), found an issue - server crashes sometimes. It mostly happened when more than one players online

With permanent quantity of bots is no problems

As commented in code, there is could be timeout reason.

Please do not merge this PR until testing on YGGverse server will be completed.
But this one commit would be included here.

### DIFF
--- a/dlls/client.cpp
+++ b/dlls/client.cpp
@@ -1194,7 +1194,7 @@ void StartFrame( void )
 
 				// have to delay here or engine gives "Tried to write to
 				// uninitialized sizebuf_t" error and crashes...
-				pause_time = gpGlobals->time + 1.0;
+				pause_time = gpGlobals->time + 1;
 				break;
 			}
 			else if( strcmp( cmd, "botskill" ) == 0 )


a1batross:
Try to run engine with sanitizers enabled. 

It's `-T sanitize` option in `waf configure`.

a1batross:
It might as well be the bug in the engine. =/

dlls/client.cpp
```diff
@@ -1194,7 +1194,7 @@ void StartFrame( void )
 
 				// have to delay here or engine gives "Tried to write to
 				// uninitialized sizebuf_t" error and crashes...
-				pause_time = gpGlobals->time + 1;
```
nekonomicon, line 1197: Currently, you changed really nothing exclude how compiler optimizes operations with digit literal promotion:
old variant will be equal to:
```
pause_time = gpGlobals->time + 1.0f;
```
but new variant will be equal to:
```
pause_time = (float)((double)gpGlobals->time + 1.0);
```

nekonomicon (approved):

dlls/client.cpp
nekonomicon: You can exclude this commit via `git rebase`.

nekonomicon: Anyway I recommend to use integer or float literals(.0f) if type of used variables is float.
For differrent types in C and C++ works promotion rules and compiler can't just replace float and double literals.
Double literals(.0) eats more memory and double-to-float/float-to double casts is not cheap for CPU.

PR #427, opened 2024-01-18
HL 25th anniversary pushable fix.

nekonomicon:


FreeSlave:
I would keep our pushablemode cvar.
Make it so 0 means the current anniversary behavior, and -1 is legacy behavior. What do you think?

PR #428, opened 2024-01-20
Port func_vehicle implementation from cs16-client/regamedll_cs.

nekonomicon:


dlls/vehicle.cpp
```diff
@@ -0,0 +1,996 @@
+#include "extdll.h"
```
a1batross, line 1: Maybe credit ReGameDLL_CS developers here?

nekonomicon: I'm not sure how it may look.

a1batross:
Yeah, they don't have anything on this file...

FreeSlave:
I tried the branch with this map https://twhl.info/vault/view/6815
I can't turn the vehicle for some reason, only move forward.
Also the sound doesn't stop after I stop the vehicle, but it behaves so in HL as well.

And regamedll claims to be GPL so I dunno how it's gonna work with hlsdk license.

FreeSlave:
Found the proble with turnings: https://github.com/FWGS/hlsdk-portable/blob/2aebf12b1f18cbd695765c918bb36b1341783cd4/dlls/player.cpp#L1905
If stops train control when player strafes.

nekonomicon:
@FreeSlave try the latest commit.

> And regamedll claims to be GPL so I dunno how it's gonna work with hlsdk license.

It has exception, but IMHO It's very unclear.
Anyway we can modify this code when valve released SDK update.

FreeSlave:
Turn controls work fine after the latest change

N7P0L3ON:
@nekonomicon any idea how to properly credit the player when he runs over / kills someone? The current implementation just takes 1 frag out of the one driving the func_vehicle for each kill... that might be desirable if you have it in Counter Strike and you're running over a teammate.. but what about if we push this to Half-Life? There it's free for all... and maybe for shits and giggles you need to get a point for fragging someone say with a forklift or other vehicle?

nekonomicon:
@N7P0L3ON Why not ask Valve about that?
Also, Half-Life has teamplay gamemode.

N7P0L3ON:
@nekonomicon Thanks for the good suggestion. We got it fixed already.

PR #431, opened 2024-01-26
HL 25th anniversary update rpg rocket changes.

nekonomicon:


mittorn:
Explode is non-virtual, will this work as supposed?

nekonomicon:
> Explode is non-virtual, will this work as supposed?

Fixed.

PR #436, opened 2024-02-10
Android CI support

Velaron:
I think APKs should be signed with something though, as without a signature they are hard to install

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -147,4 +147,4 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: hlsdk-${{ steps.extract_branch.outputs.branch }}-android
-          path: android/app/build/outputs/apk/debug/*
+          path: android/app/build/outputs/apk/debug/*.apk
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -138,6 +138,11 @@ jobs:
         env:
           ANDROID_NDK_ROOT: /usr/local/lib/android/sdk/ndk/26.1.10909125
 
+      - name: Extract branch name
+        shell: bash
+        run: echo "branch=$(echo ${GITHUB_REF#refs/heads/} | tr '/' '_')" >> $GITHUB_OUTPUT
+        id: extract_branch
+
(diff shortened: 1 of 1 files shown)


nekonomicon:
> I think APKs should be signed with something though, as without a signature they are hard to install

Nah, I'm against distributing random key signed builds.
And you probably didn't understand me.
CI needed for release builds.
Our hlsdk fork is stable enough, so there no reason to make debug builds.
And IMHO good testing environment - any linux distro compiled with clang or even better *BSD.

Velaron:
I don't see anything wrong with signing APKs, without a signature users will have to sign it by themselves or use ADB, which would require a PC and some knowledge. Then they will distribute these signed APKs through random chats and websites, because that is what happens. You can use your oersonal debug.keystore or official keys I use in the Play Store.
As for the second point, in my experience Android has a lot of unique bugs and crashes that I can't reproduce anywhere, but debug was just a typo.

a1batross:
Like I did a thousand times, publish a keystore somewhere (maybe even in
the repo itself) and use it in CI.

Those builds shouldn't be trusted anyway, but at least it solves the
problem of unsigned APKs and random key signed APKs (which disallow upgrade)

Velaron:
The question is, do I add my debug signautre I use in some repos, or the release signature used in Google Play?

FreeSlave:
Github actions support secrets. Why not use them if you need some private key?

Velaron:
Yeah, that's what I plan to use for GP signatures

a1batross:
>The question is, do I add my debug signature I use in some repos, or the release signature used in Google Play?

It doesn't matter to me. You can as well just make a standalone repo (as it might be shared with other projects, like engine) and fetch it using wget or curl.

a1batross:
> Github actions support secrets. Why not use them if you need some private key?

this is good for storing passwords. If you encode a keystore into a text, it probably needs to be stored somewhere. I don't think apksigner knows how to read from stdin.

Velaron:
I upload the key as base64 and then it's decoded back into a file, I already use this method on my mods

PR #440, opened 2024-02-23
server: export physic interface init function to properly detect Xash3D engine

a1batross:


jeefo:
And what if we're have metamod between gamedll and engine ?:)

PR #443, opened 2024-03-13
Use SDL2 mouse input on Windows

FreeSlave:
Dynamically load SDL2 on Windows and use its functions for mouse handling. If SDL2 is not available (i.e. the mod is running on pre-SDL2 version of Half-Life), it fallbacks to the pre-SDL2 mouse input.
`USE_SDL2` macro is removed and replaced with dynamic checks for the presence of `sdl2Lib`.

Without `SDL_SetRelativeMouseMode` and `SDL_GetRelativeMouseState` the raw input doesn't really seem to work, so I decided it could be useful to have this option working in this hlsdk. But we need to figure out the reliable way to check if raw input works without relying on "feelings".

Although it works, you can consider this merge request as WIP, because I left joystick input the same as in pre-SDL2 Half-Life versions on Windows regardless of whether SDL2 is available or not. I don't have a joystick currently, so I can't test anything related to it. Someone else can change and test such things and add commits to this PR.

I also haven't tested it on Xash3D-FWGS yet (just to make sure it doesn't cause any troubles while building without GoldSource input support) and on the vanilla Xash3D.

I also added SDL2 headers (copied from valve's repo external/ subdirectory), so the user is not required to install SDL2 headers on Windows. On Linux you still need to download a dev package to SDL2. But we can rely on the headers bundled in the repository for Linux build too.

Thoughts:
We probably can remove some SDL2 .h files to minimize their presence in the repo.


FreeSlave:
Deleted redundant SDL2 headers

a1batross:
I wonder what we're going to do when SDL3 will come out. Will Valve upgrade GoldSrc to it? I hope not...

Though for us, it has some interesting features I would like to leverage in the engine, mostly supporting multi-monitor setups. But I don't think we will migrate to it anytime soon, as SDL2 will continue to exist and packaged on modern systems, like SDL 1.2 is still required by some software.

a1batross:
I think there are few more redundant headers:

* SDL_config_iphoneos.h: not supported by GoldSrc
* SDL_config_android.h: not supported by GoldSrc
* SDL_config_psp.h: not supported by GoldSrc
* SDL_config_nintendods.h: not referenced by anything
* SDL_config_pandora.h: not referenced by anything
* SDL_config_wiz.h: not referenced by anything

FreeSlave:
> I think there are few more redundant headers:
> 
> * SDL_config_iphoneos.h: not supported by GoldSrc
> * SDL_config_android.h: not supported by GoldSrc
> * SDL_config_psp.h: not supported by GoldSrc
> * SDL_config_nintendods.h: not referenced by anything
> * SDL_config_pandora.h: not referenced by anything
> * SDL_config_wiz.h: not referenced by anything

Removed them as well

FreeSlave:
I pushed changes that makes the game use SDL2 joystick code on Windows as well.
I don't have a joystick myself, so it all needs to be re-tested (on Linux too).

If you want to check if winapi joystick still works, you can change how `UseSDL2Joystick` to always return false (just for a test).
We could also provide a cvar to let user decide what joystick input they prefer (e.g. `joy_prefer_winapi`). But I think it's better if someone with joystick device makes these changes.

FreeSlave:
@a1batross any progress with testing on your side? Can it be merged?

a1batross:
I will look into that soon.

a1batross:
I forgot. What exactly I should look for?

It builds and runs on Linux versions of GoldSrc and Xash3D just fine.

FreeSlave:
@a1batross 
* Check that raw input now works correctly on Windows on GoldSource.
* Check if joystick still works on Windows and Linux on GoldSource.
* Check if joystick still works on pre-SDL2 version of Half-Life.

a1batross:
Everything works. LGTM

FreeSlave:
@nekonomicon merge it?

PR #448, opened 2024-04-05
Register cvars containing information about current build

a1batross:


nekonomicon:
Which reason to provide pointers to global variables in setter way?

erfty125:
> Which reason to provide pointers to global variables in setter way?
> Anyway I see useless complexity here.
> 

i have only 2 states:
i dont use getters/setters 
i use getters/setters everywhere
pattern chud

a1batross:
Can you review it once more @nekonomicon?

PR #449, opened 2024-04-14
Rename EFLAG_MONSTER to EFLAG_FLESH_SOUND

a1batross:
This is mentioned in default delta.lst shipped with HL25:

```
	DEFINE_DELTA( eflags, DT_INTEGER, 2, 1.0 ), // JoshA: Changed to 2 for EFLAG_FLESH_SOUND to fix flesh sounds on entities. If a mod ships updated DLLs and not this file, they just won't have that fix.
```


a1batross:
Just to match HLSDK 2.5 when it releases.

If it releases, of course.

PR #451, opened 2024-05-08
Fix func_door_rotating with 'Starts Open' spawnflag

FreeSlave:
There's an error in the door angles swap which leads to the rotating door with `Starts open` spawnflag not moving after it was "closed". The Toggle door won't respond to the trigger and non-Toggle door won't return to the position.

This is probably dangerous to merge as some existing maps may rely on this behavior.
I leave this PR just to document the bug.

### DIFF
--- a/dlls/doors.cpp
+++ b/dlls/doors.cpp
@@ -892,7 +892,7 @@ void CRotDoor::Spawn( void )
 	{	
 		// swap pos1 and pos2, put door at pos2, invert movement direction
 		pev->angles = m_vecAngle2;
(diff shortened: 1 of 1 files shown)


mittorn:
Maybe such things should be in some "bugcomp" mode under macro?
Authors of new mods do not need to rely on specific half-life behaviour

nekonomicon:
> Maybe such things should be in some "bugcomp" mode under macro?

Yeap, will be better to place new code under macro instead of stillborn PR.

PR #458, opened 2024-07-01
Fix #238

FreeSlave:
My proposal for fixing the issue [#238](https://github.com/twhl-community/halflife-updated/issues/238)

Forcefully reset the frame to 0 when animation is no longer played to avoid having the situation when the grenade is stuck on the 'flying' frame.

Note: the frame reset is visible of course, but not too noticeable during the game.

Alternatively we can just revert the commit that made grenades animated in the first place.


FreeSlave:
I realized I put an issue number from another repo in the commit...

PR #464, opened 2024-08-15
Fix killing tentacle in Blast Pit

a1batross:
Related https://github.com/FWGS/xash3d-fwgs/issues/1771

### DIFF
--- a/dlls/util.cpp
+++ b/dlls/util.cpp
@@ -2217,7 +2217,7 @@ int CRestore::ReadField( void *pBaseData, TYPEDESCRIPTION *pFields, int fieldCou
 						break;
 					case FIELD_EHANDLE:
 						// Input and Output sizes are different!
-						pInputData = (char*)pData + j * gInputSizes[pTest->fieldType];
+						pInputData = (char*)pData + j * gSizes[pTest->fieldType];
 						entityIndex = *(int *)pInputData;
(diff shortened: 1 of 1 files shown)


FreeSlave:
The commit in hlsdk-portable https://github.com/FWGS/hlsdk-portable/commit/f58544a1568f6abbbe28bf3c5c94b5b967e4f69c didn't change the gSizes to sizeof(EHANDLE) for FIELD_EHANDLE. Could it be the culprit?

FreeSlave:
I've made a simple test map with multisource (it uses the array of  EHANDLES).
[ms_test.zip](https://github.com/user-attachments/files/16624866/ms_test.zip)

Press the first and the third button. Then save the game, reload and press the second button.
The fourth button should become available for press now. But in hlsdk-portable it still denies the input.

a1batross:
Wouldn't it break compatibility with normal Half-Life save files?

I didn't try your solution yet...

FreeSlave:
It's for reading into memory. So it shouldn't. `gInpusSizes` define the sizes in the save file. `gSizes` define the sizes of variables in the memory.

a1batross:
Probably superseded by https://github.com/FWGS/hlsdk-portable/pull/467 now

But I'm not sure if we should change such things in the first place, if they tend to break existing games.

PR #473, opened 2024-10-17
Fix unreferenced variable.

malortie:
`pMonster` on line 1067 was unreferenced.

The code has been changed to keep the first declaration so that it follows the same convention as in the other functions (variables declared at the top).


a1batross:
Maybe reduce the scope of this variable?

Just keep it as was but remove `CBaseMonster *pMonster` declaration at the top of the function.

a1batross (approved):

PR #477, opened 2024-10-24
Fix sv_game_build_branch referencing the wrong string

FreeSlave:


a1batross (approved):
Oops, that was my mistake. Good job!

PR #479, opened 2024-10-24
Build github artifacts using Ninja

FreeSlave:
Building this sdk with Makefiles takes ~44 seconds whereas building it with Ninja takes ~18 seconds  (for clang build it's 24 seconds vs 10 seconds).

We also need to look at the ways of improving the build times on Windows.

Another possible speedup for Linux is using podman to install steam runtime (podman is preinstallted on github Ubuntu images, unlike schroot), but then the container data needs to be cached.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,13 +52,13 @@ jobs:
     - name: Build on Linux
       if: startsWith(matrix.os, 'ubuntu')
       run: |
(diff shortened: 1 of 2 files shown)


a1batross:
Of course, it will take 44 seconds vs 18 seconds because Ninja by default runs as multiple concurrent jobs and GMake for some stupid historic/unix legacy reasons doesn't, even if you invoke it through `cmake --build`. With the same amount of parallel jobs, Ninja usually wins like few seconds from GMake.

Anyway, good catch. Ninja is nice for CI anyway.

As for Windows, I believe that running CL.EXE with multiple source files doesn't make it use multiple CPU cores. There, however, was a flag for that. It's also possible to run Ninja on Windows too. It doesn't make much sense to generate VS project on CI.

a1batross (approved):

PR #481, opened 2024-10-29
Copy vgui.so to the base build directory in github actions

FreeSlave:
After switching to Ninja the vgui autobuild is broken because ninja uses a different current directory.
I kept copying vgui.so to the cl_dll/ intact just in case someone changes generator back to Makefiles.


a1batross:
I think it's only required for steamrt build because the issue with vgui.so is being linked by full path is an ancient binutils bug.

FreeSlave:
Maybe. But github actions are currently using steamrt.
This allows to build libraries that are surely compatible with GoldSource on Linux.
As far as I remember I was having problems with some gcc/g++ versions producing incompatible binaries.

a1batross:
Wait, everything uses steamrt? I thought `manual.yml` just uses whatever is on CI machine.

FreeSlave:
Can we merge it? Users are complaining already! :)

a1batross (approved):

PR #484, opened 2024-11-21
Use latest scout steam runtime

FreeSlave:
I noticed the steam scout runtime sdk from 2021 is not available anymore. Valve probably deleted old archives. Try using latest-public-beta instead.

If you're interested here's the list of snapshots: https://repo.steampowered.com/steamrt-images-scout/snapshots/

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,6 +20,7 @@ jobs:
     env:
       CC: ${{ matrix.cc }}
       CXX: ${{ matrix.cxx }}
+      STEAMRT_SNAPSHOT: latest-steam-client-general-availability
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


FreeSlave:
huh, that's weird. Github used the cached tar.gz this time.
But for some reason it didn't use the cache when building for https://github.com/FWGS/hlsdk-portable/commit/387c266c05cf041c38f2296d74dec6e1ced6eb5a

Anyway, the .tar.gz from 2021 is gone. And we need to update the link.

a1batross:
You can manually purge caches in https://github.com/FWGS/hlsdk-portable/actions/caches. 

Also, I've looked into scout steamrt recently. It actually provides newer compilers _and_ newer binutils. So maybe we could use that to avoid that vgui.so linking issue.

FreeSlave:
> huh, that's weird. Github used the cached tar.gz this time.

Nevermind. It did download the steam runtime first time before I force-pushed with another commit name.

I'm making it more configurable by letting to choose the snapshot.

FreeSlave:
Btw, I'm not sure what's difference between  latest-public-beta, latest-steam-client-general-availability and latest-steam-client-public-beta. They all have the same BUILD_ID.txt currently.
Maybe they may become different when Valve is preparing the new runtime.

[Here](latest-steam-client-general-availability) it explains that latest-steam-client-general-availability should be used usually.

FreeSlave:
I set it to latest-steam-client-general-availability and squashed the commits.
Please review again.

a1batross (approved):

FreeSlave:
@nekonomicon merge?

PR #485, opened 2024-11-23
Fix choosing the earth target for trigger_xen_return

FreeSlave:
In Opposing Force the target to travel back to Earth is the one closest to the player's origin before he travelled to Xen.


FreeSlave:
Ah, need https://github.com/FWGS/hlsdk-portable/pull/484 to be merged first. Including merging into mod branches.

PR #488, opened 2024-11-25
Delete .lib files before uploading artifact for Windows

### LINKED ISSUE #486: Artifacts for Win32 wrongly include .lib files
Engine or users don't need .lib files for neither client or server DLLs.

These files shouldn't be shipped.

FreeSlave:
Fix #486 

This also changes deleting `dlls/hl.lib` to deleting `dlls/*.lib` in manual build to deal with cases when server dll is not `hl` (e.g. `opfor`). Of course I could instead extract the `SERVER_LIBRARY_NAME` from cmake, but it's easier this way.


a1batross:
hm, I think CMake should have the option what to install...

FreeSlave:
Superseded by #489

PR #490, opened 2024-11-28
Client: fix draw menu

Elinsrc:
Do not draw the old menu of the previous server when you connect to another server


a1batross:
Looks fair to me. It's unlikely that the server will preserve menu state between changelevels.

PR #491, opened 2024-11-29
Novgui voice chat support

Elinsrc:
Added support for the voice chat interface without vgui and the display of the icon in the novgui scoreboard (it is not completed because I am lazy), if necessary, you can finish and redo it as needed

### DIFF
--- a/cl_dll/CMakeLists.txt
+++ b/cl_dll/CMakeLists.txt
@@ -124,6 +124,8 @@ set (CLDLL_SOURCES
 	tri.cpp
 	util.cpp
 	view.cpp
+	voice_status.cpp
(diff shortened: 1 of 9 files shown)


SNMetamorph:
That's good idea. But I think implementation is somehow bloated. Maybe it's possible to implement such thing without modifying VGUI-related code, and consequently many of that `#ifdef` won't be needed.

Velaron:
Some files changed are just unnecessary formatting, like voice_status.cpp/h, try reverting those

Velaron:
Also there is a lot of copy-pasted code, I think it's possible to merge both files and put the modified lines under macros?

PR #505, opened 2025-01-02
Floor taken damage for player instead casting to int

FreeSlave:
Straight casting from `float` to `int` may result in the number becoming negative when the original number is positive if original `float` number is too large to be represented in `int`.

Example: the map `semonz` (distributed with Adrenaline Gamer mod) has several instances `trigger_hurt` with damage set to `"10000000000"` which can't be represented in `int`. It becomes negative which leads to increasing health in `TakeDamage`. Then in `V_CalcViewRoll` on client `pparams->health` becomes negative (probably some other cast in the chain?). Player continues to be considered alive by the server, but he has a rolled camera because the view code thinks the player is dead.

Note: ideally the `trunc` should be used instead of `floor` for positive values and `ceil` for negative ones. But it's C++11 and this repo is still C++98.

### DIFF
--- a/dlls/player.cpp
+++ b/dlls/player.cpp
@@ -486,7 +486,7 @@ int CBasePlayer::TakeDamage( entvars_t *pevInflictor, entvars_t *pevAttacker, fl
 
 	// this cast to INT is critical!!! If a player ends up with 0.5 health, the engine will get that
 	// as an int (zero) and think the player is dead! (this will incite a clientside screentilt, etc)
-	fTookDamage = CBaseMonster::TakeDamage( pevInflictor, pevAttacker, (int)flDamage, bitsDamageType );
+	fTookDamage = CBaseMonster::TakeDamage( pevInflictor, pevAttacker, flDamage >= 0.0f ? floor(flDamage) : ceil(flDamage), bitsDamageType );
 
 	// reset damage time countdown for each type of time based damage player just sustained
 	{


mittorn:
This may propogate possible NaNs widely, exploding entire server

a1batross:
NaNs should be handled elsewhere.

For example, by properly checking what gets passed to sqrt or division, etc. Not in random place like this.

FreeSlave:
> This may propogate possible NaNs widely, exploding entire server

How is it different from NaN getting into TakeDamage function of any other entity (monsters, breakables) which don't cast to integer?

mittorn:
Other entities usually does not take damage in multiplayer, in singleplayer this is not so serious issue. But there was some bug with hornetgun in gravgun branch, allwoing NaN in player health and godmode bug in multiplayer.
Damage somethimes depends on trace and on some big numbers, so getting NaN is very likely. If it was casted to int, maybe need add some NaN-friendly bounds checking

mittorn:
Good luck find all NaNs and check all sqrt's. Good point checking for NaNs is where entities interact after calculations are done, so damage functions seems to be good place. Maybe add log printing source entity classname and damage type, to make easier finding nan source.

a1batross:
> Good luck find all NaNs and check all sqrt's. Good point checking for NaNs is where entities interact after calculations are done, so damage functions seems to be good place. Maybe add log printing source entity classname and damage type, to make easier finding nan source.

FYI I did that. Twice.

FreeSlave:
Is casting NaN to integer even a defined behavior in standard? What are you relying on here?

mittorn:
It is UB, but changing to ffloor may change behaviour

PR #511, opened 2025-02-22
fix mp_bhopcap

Elinsrc:


FreeSlave:
I think "cap" in this context means "capped", which means when bhopcap is non-zero, the bunnyhop is disabled.

It's not perfect name and I would prefer cvar with positive meaning, i.e. when cvar value is non-zero, it should mean the feature is enabled.

Elinsrc:
does that mean there is a typo in the description?
image

nekonomicon:
It was made for compatibility with miniag.
Wrong description fixed in https://github.com/FWGS/hlsdk-portable/commit/33652de986f6c0e97def9d1833326e3f10bd1adf

PR #519, opened 2025-04-14
Fix alien grunt tracer effect end position

FreeSlave:
The problem:
Alien grunt code sends the direction vector as the second vector in TE_TRACER message. But the message expects both vectors to represent positions.

This explains random tracers you could see while being near the world origin when alien grunts fight human grunts.

The test map to reproduce the issue on vanilla Half-Life: [agrunt_test.zip](https://github.com/user-attachments/files/19734530/agrunt_test.zip)


a1batross:
Looks correct to me!

PR #538, opened 2025-06-04
Fix compiler error for enumeration constants

kairusds:


a1batross:
> This is wrong, looks like error caused by compiler parameters. Perhaps it uses C99 standard by default, and it does have `true` and `false` keywords, unlike C89.

No, it's actually C23 that made them keywords and modern versions of GCC already default to it. To me, the qboolean definition that depends on the language and standard version is a bit of a nonsense and I think the safest solution would be always defining qboolean as int and only providing false and true for C below C23. That's actually what I did for the engine: https://github.com/FWGS/xash3d-fwgs/commit/de196b1c588d896d24ed402b6046d57370540611

common/const.h
```diff
@@ -743,11 +743,11 @@ typedef unsigned short	word;
 #undef true
 #undef false
 
-#if !__cplusplus
-typedef enum { false, true }	qboolean;
-#else 
-typedef int qboolean;
+// https://github.com/FWGS/xash3d-fwgs/commit/de196b1c588d896d24ed402b6046d57370540611
+#if !__cplusplus &&  __STDC_VERSION__ < 202311L
+enum { false, true }
```
nekonomicon, line 748: It seems there missing semicolon.

nekonomicon (changes requested):

SNMetamorph:
LOL

nekonomicon:
Yep.
All Good.
Thank you.

PR #541, opened 2025-06-22
Use Ninja generator for Windows autobuilds

FreeSlave:
Building with ninja is ~2x faster than with msbuild.

Ninja doesn't support platform specification, so I'm using the github action to switch to x86 developer prompt.

I tested vgui and non-vgui build. I didn't test 64-bit one.


a1batross:
You might squeeze a bit more with amd64_x86 compiler. It's a 64-bit compiler targetting 32-bit code.

FreeSlave:
> You might squeeze a bit more with amd64_x86 compiler. It's a 64-bit compiler targetting 32-bit code.

Are you referring to some custom compiler toolset? Or some other options to pass to cmake to avoid setting the command prompt?

PR #553, opened 2025-11-02
Fix Xen Attack not working properly.

weimzh:
The root cause is m_bSlaveCoop was set in CWorld::KeyValue but got immediately resetted in CWorld::Spawn.  This fixes the issue.


a1batross (approved):
My mistake, sorry

PR #557, opened 2026-02-02
Big endian support

IntriguingTiles:
Figured I'd start with something relatively small. This should be everything that needs to be byteswapped in hlsdk.

I've also added definitions for the s390x architecture which is the only big endian architecture that has official support from debian and some other big distros, which might be useful for CI.

### DIFF
--- a/dlls/nodes.cpp
+++ b/dlls/nodes.cpp
@@ -24,6 +24,7 @@
 #include	"nodes_compat.h"
 #include	"animation.h"
 #include	"doors.h"
+#include	"byteswap.h"
 
 #define	HULL_STEP_SIZE 16// how far the test hull moves on each step
 #define	NODE_HEIGHT	8	// how high to lift nodes off the ground after we drop them all (make stair/ramp mapping easier)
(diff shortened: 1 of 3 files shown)


cl_dll/StudioModelRenderer.cpp
```diff
@@ -404,8 +406,29 @@ mstudioanim_t *CStudioModelRenderer::StudioGetAnim( model_t *m_pSubModel, mstudi
...
+		mstudioanim_t *panim;
+		mstudioseqdesc_t *pseqdesc2 = (mstudioseqdesc_t *)( (byte *)m_pStudioHeader + m_pStudioHeader->seqindex );
 		gEngfuncs.Con_DPrintf("loading %s\n", pseqgroup->name );
 		IEngineStudio.LoadCacheFile( pseqgroup->name, (struct cache_user_s *)&paSequences[pseqdesc->seqgroup] );
+		
+		for ( int i = 0; i < m_pStudioHeader->numseq; i++ )
+		{
+			if ( pseqdesc2[i].seqgroup == pseqdesc->seqgroup )
+			{
+				panim = (mstudioanim_t *)( (byte *)paSequences[pseqdesc->seqgroup].data + pseqdesc2[i].animindex );
```
a1batross: Let's reduce `panim` scope to this block.

PR #558, opened 2026-02-06
Add termux build script

erfty125:
Goofy script, but it works

### DIFF
--- a/android/termuxbuild.sh
+++ b/android/termuxbuild.sh
@@ -1,5 +1,5 @@
 #0. install build tools:
-#   apt update && apt install nano aapt openjdk-21 dx apksigner cmake
+#   apt update && apt install aapt openjdk-21 dx apksigner cmake
 #1. install https://github.com/lzhiyong/termux-ndk( to compile libs ),
 #  set path to this dir in NDKDIR variable
 #2. get android.jar file, for example:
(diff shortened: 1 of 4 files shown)


android/termuxbuild.sh
```diff
@@ -0,0 +1,114 @@
...
+      -B build
+cd android
+fi
+
+cp ${WPATH}/AndroidManifest.xml build
+sed -i '/<manifest/apackage="su.xash.hlsdk"' build/AndroidManifest.xml
+sed -i '/<\/queries>/a<uses-sdk android:minSdkVersion="3" android:targetSdkVersion="34"\/>' build/AndroidManifest.xml
+#default manifest file not containing this lines, so we just copying into it
+
+aapt package ${AGRF}
```
a1batross: This whole method is somewhat too stupid.

While you can obviously do that, ever slight change in the code like adding some external dependencies, would require changes in this script

Better figure out how to call gradle instead.

a1batross (changes requested):

android/termuxbuild.sh
erfty125: I agree with you, but gradle doesnt work in  termux "out of box", you have to do many goofy tricks to make gradle work on termux + gradle has many bloat dependencies, no one need that for building dummy xash mod launcher

a1batross: You can't assume that nobody needs it because this is just a sample.

That's honestly just my opinion. Maybe @nekonomicon or @Velaron has some different thoughts on this matter.

PR #562, opened 2026-03-16
Byteswap routines for unaligned data on mips-irix

Xav101:
Now that PR https://github.com/FWGS/hlsdk-portable/pull/557 is merged for big endian support, these are the specific changes needed to make it work properly on IRIX.

Unlike some other OSes on RISC platforms that required alignment, IRIX does not trap and emulate unaligned accesses. Therefore, we need to handle any cases of unaligned access specifically in software. Therefore, this patch adds a new byteswap method specifically for accessing unaligned data. We don't want to take the unaligned read/unaligned write penalty on every single `LittleToHost` method. In the majority of cases the data seems to be well-behaved, and I've only used the unaligned variant in instances where I've encountered a SIGBUS crash.

As-is this patch only implements `ULittleToHost`/`ULittleToHostSW` on big endian platforms, with the implementation simply not doing anything on little endian platforms. While it's possible that a little endian platform that requires aligned data. any ports to such platforms seem to be entirely theoretical at this point. Additionally it's worth noting that a number of the more popular little-endian RISC platforms already supported handling unaligned accesses. (Tru64 I think has a handler but generates a configurable log message, and all of the Windows NT RISC ports had handlers.) The patch also has no unaligned equivalent for `BigToHost`/`BigToHostSW` because I haven't found any instances of one being needed. If one does crop up at some point however it would be simple enough to add it later.

### DIFF
--- a/common/byteswap.h
+++ b/common/byteswap.h
@@ -44,16 +44,50 @@ inline float Byteswap( float n )
 	return SwapFloat( n );
 }
 
+template<typename T>
+inline T UByteswap( T& n )
+{
+	T u;
+	memcpy(&u, &n, sizeof(u) );
+	switch ( sizeof(u) )
+	{
+		case 2: return Swap16( u );
+		case 4: return Swap32( u );
+	}
+}
+
+template<>
+inline float UByteswap( float& n )
+{
+	float u;
+	memcpy(&u, &n, sizeof(u) );
+	return SwapFloat( u );
+}
+
+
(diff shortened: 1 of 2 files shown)


Xav101:
Just pushed a new version of the patch that uses references and ditches the `std::remove_pointer` requirement. I playtested this for the common issues I've encountered and it seems fine. I just played a new save up through the start of office complex and it seems to work fine.

PR #571, opened 2026-05-01
readme: fix remaining hlsdk-xash3d references.

malortie:
This changes the remaining references of hlsdk-xash3d (previous name) to hlsdk-portable (current name).


nekonomicon:
Thanks!

PR #585, opened 2026-06-04
Cracklife

znQci:


a1batross:
@nekonomicon do we want this considering engine can just download libraries from the Internet?

Velaron:
Completely useless, if not right now then in the future

nekonomicon:
@a1batross nope.

PR #592, opened 2026-08-05
freevgui

a1batross:
Not gonna build as freevgui repo is not public yet, though FWGS members can access it.

I had to restore original vgui headers due to vgui_support not including them in vgui-dev anymore. In fact, I kinda want to retire vgui_support code and instead include Surface implementation in FreeVGUI itself, so that might get removed too, as the plan for "unsupported-by-Valve" platforms is to statically link VGUI so we don't have to handle possible C++ ABI incompatibilities on less stable platforms.

I tried to preserve fake_vgui semantics contributed by @FreeSlave but I'm not sure if we need this feature at all, as it was intended to be used on platforms supported by Xash3D FWGS that weren't including VGUI support by default. Now that it is going change, maybe we don't need this anymore.


nekonomicon:
> I tried to preserve fake_vgui semantics contributed by @FreeSlave but I'm not sure if we need this feature at all, as it was intended to be used on platforms supported by Xash3D FWGS that weren't including VGUI support by default. Now that it is going change, maybe we don't need this anymore.

It was temporary solution to fix 2 annoying bugs under GoldSource.
And it was actually when @mittorn removed vgui code.
If we have own portable solution there no reason to compile HLSDK without vgui and voice manager anymore.

a1batross:
Yeah, I agree with you here. We can clean up quite a lot of code, and port
more mods.

FreeSlave:
The main point is to provide game libraries that can be shared between GoldSource and Xash3D-FWGS (on x86 Linux and Windows).

If the same client can run on both (Xash3D-FWGS with freevgui and GoldSource with original vgui), then sure, there will be no need in building two different versions. (I would still keep non-vgui scoreboard and motd, at least as an option, because 25th anniversary GoldSource breaks vgui on Linux if the game runs not on the leftmost monitor, but it's another topic)

I would test it out, but I'm not sure how to build freevgui. Need build instructions.

a1batross:
Yeah, freevgui is ABI compatible with the original, that's the whole point.

I will finish it and make it public pretty soon, it's almost 95% done.

a1batross:
@nekonomicon it's pretty much done and seems to just work. I didn't cleanup novgui code, just in case if it will be ever useful.

I also received a question from @TheKingFireS about missing voicemgr, now that we have freevgui everywhere, I think we should drop USE_VOICEMGR build time option as well, though that is probably out of scope for this PR.

nekonomicon:
LGTM.

PR #599, opened 2026-09-08
cmake: Fix ios naming

ksagameng2:
Messed it up in last pr


a1batross:
Can you also update library_suffix.py? I know it's probably not going to be used, just to keep them in sync with each other.

ksagameng2:
Done, lmk if there's a mistake because I can't test it

scripts/waifulib/library_naming.py
```diff
@@ -184,7 +186,7 @@ def configure(conf):
 	node = conf.bldnode.make_node('true_postfix.txt')
 	node.write('%s-%s' % (buildos, buildarch))
 
-	if not conf.env.XASH_ANDROID and (conf.env.XASH_WIN32 or conf.env.XASH_LINUX or conf.env.XASH_APPLE):
+	if not conf.env.XASH_ANDROID and not conf.env.XASH_IOS and (conf.env.XASH_WIN32 or conf.env.XASH_LINUX or conf.env.XASH_APPLE):
```
a1batross, line 189: Is it really needed for iOS?

Do iOS libraries should have _ios_arm64 suffix? If so, then this check is not needed.

nekonomicon: @a1batross XASH_APPLE?

ksagameng2: > Is it really needed for iOS?
> 
> Do iOS libraries should have _ios_arm64 suffix? If so, then this check is not needed.

Yes they should, but XASH_APPLE is also set for ios so this check is needed (basically same as android having XASH_LINUX set but still requiring different name)

a1batross: This is honestly a bug on my part...

PR #600, opened 2026-09-10
Add cl_motd_vgui cvar

FreeSlave:
Add cl_motd_vgui cvar to control whether to show vgui or non-vgui MOTD window


FreeSlave:
Oops, forgot to remove some USE_NOVGUI_MOTD ifdefs. Squash it before merging.

